PR #45, opened 2014-12-21
Fix Google authentication flow according to Google documentation

cguy:
- Fixed IDE parsing errors (trailing commas)
- Fixed Google authentication flow according to Google documentation : https://developers.google.com/accounts/docs/OAuth2WebServer#handlingtheresponse

However, you may test your current implementation before include mine. For your information, my Google "project" has been created few hours ago.

The issue was, in my opinion, because Stormpath didn't follow the Google authentication for the server side but the client one. So a step was missing (convert oauth2 code into access token).

### DIFF
--- a/lib/controllers.js
+++ b/lib/controllers.js
@@ -2,6 +2,7 @@
 
 
 var async = require('async');
+var Client = require('node-rest-client').Client;
 
 var forms = require('./forms');
 var helpers = require('./helpers');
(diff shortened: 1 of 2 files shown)


rdegges:
Hi @cguy, sorry it's taken so long to get back to you on this.

I'm not really sure I understand what this is doing, exactly. The way our existing stuff is working works for me out of the box. EG: We create a Google login button, request scopes / permissions, and the user is then redirected back to the application with an access token that is verified.

rdegges:
Closing this for now until I hear back!

PR #56, opened 2015-03-11
Adds support for passing in a caching client directly

jeffkole:
We found ourselves wanting to use a pre-configured Redis connection, specifically for the authentication support.  I believe the solution will work for a pre-configured Memcached client, too.

I reached the application limit when running the full suite of tests.  If you have a workaround, I am happy to run the tests locally to verify this changeset.

### DIFF
--- a/lib/settings.js
+++ b/lib/settings.js
@@ -68,6 +68,7 @@ module.exports.init = function(app, opts) {
   app.set('stormpathCacheTTL', opts.cacheTTL || parseInt(process.env.STORMPATH_CACHE_TTL) || 300);
   app.set('stormpathCacheTTI', opts.cacheTTI || parseInt(process.env.STORMPATH_CACHE_TTI) || 300);
   app.set('stormpathCacheOptions', opts.cacheOptions || JSON.parse(process.env.STORMPATH_CACHE_OPTIONS || '{}') || {});
+  app.set('stormpathCacheClient', opts.cacheClient || undefined);
 
   // Oauth configuration.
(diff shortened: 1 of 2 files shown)


jeffkole:
The build failed because the `apiKey.id` was not set.  Is that something I should have control over in Travis?

lhazlewood:
The CI build uses [Travis CI's secure environment variables](http://docs.travis-ci.com/user/environment-variables/).  Here's our `travis.yml`: https://github.com/stormpath/stormpath-express/blob/master/.travis.yml

I believe this sets `STORMPATH_API_KEY_ID` and `STORMPATH_API_KEY_SECRET` environment variables with actual credentials and those are used to construct the Client and perform integration tests.

lhazlewood:
Ah, I was explaining the process for the SDK project - not sure if `stormpath-express` does the exact same thing, although I would _presume_ so.  @rdegges can you please confirm?

rdegges:
Indeed -- we only supply the API key for the master/develop branches for security. But I'm happy to test it manually ^^

jeffkole:
Any update on the manual testing?

CannibalVox:
Has there been any motion on this?  Without this PR, free Redis plans on Heroku cannot be used with Stormpath, because Stormpath uses 10 connections on its own per dyno and won't even share with the rest of the application.

rdegges:
Hey @CannibalVox I'm merging this in right now! I've been super backlogged with another urgent Stormpath issue for the past couple of weeks, sorry about the slow turnaround!

Really appreciate everyone in this thread for the productive patch / comments <33 

Ya'll are the best.

CannibalVox:
Thanks a ton!

Issue #35, opened 2014-11-13
Create Account - Confirm Password


goldnmean:
I need an option of being able to add a Confirm Password field beneath the Password field on the default Create Account form.  This Confirm Password field exists on the Create Account form in the Admin Console.

rdegges:
I'll get this into the library soon =) Thanks all for the feature votes! I was initially against this <3

rdegges:
This is officially available now. You can use the new `enableConfirmPassword` and `requireConfirmPassword` settings to enable this functionality!

sholladay:
This is great, thanks @rdegges !

alexwainger:
@rdegges Could you update the documentation here: https://docs.stormpath.com/java/spring-boot-web/registration.html#form-fields

This is a great add, but i'm not totally sure how to go about using it.

markzero:
This functionality is removed now?

robertjd:
Doh!  You're totally right, this got lost in the 2.x upgrade.  I've created #257 to track this.  I'll try to prio this, can you let me know if this is blocking you?  Thanks.

markzero:
I continued without password repeat, maybe won't need it after all, but would be nice to have it as it's very common.

PR #58, opened 2015-03-31
Added enableConfirmPassword for registration form.

CannibalVox:
Completes #35

### DIFF
--- a/lib/controllers.js
+++ b/lib/controllers.js
@@ -14,6 +14,7 @@ var FIELDS = {
   'surname': 'Last Name',
   'email': 'Email',
   'password': 'Password',
+  'confirmPassword' : 'Confirm Password',
 };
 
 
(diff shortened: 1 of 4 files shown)


CannibalVox:
I don't understand why travis is linked to github if travis cannot successfully build from github.

rdegges:
Thanks, this looks great! I'll merge this in once I wrap up my current project.

For the record: the reason PR's aren't running on travis is because I've got my personal Stormpath API credentials stored in Travis CI. Anyone could submit a PR that basically says: console.log(process.env.STORMPATH_API_KEY_ID) to view my credentials -- that's why travis is configured to ONLY run tests on the master / develop branches, since we control those directly.

It's annoying, I know -- but it's for security ^^

rdegges:
This is included in the latest release -- but I disabled it by default. I still think it's best to have this as an optional field, but not on by default.

I think the way UX is going (into the future) is to require as little info as possible.

PR #60, opened 2015-04-01
Retains query parameters upon login redirect

### LINKED ISSUE #59: `authenticationRequired` middleware loses query parameters of original URL
On [line 180 of authentication.js](https://github.com/stormpath/stormpath-express/blob/master/lib/authentication.js#L180), the query parameters are explicitly stripped off from the URL before redirecting to the login page.

```
var url = req.app.get('stormpathLoginUrl') + '?next=' + encodeURIComponent(req.originalUrl.split('?')[0]);
```

This means that if you navigate to a page with query parameters and are redirected to login, you will not arrive back at the original page once you complete login.  It looks intentional though.  What was the rationale?

jeffkole:
Resolves #59

### DIFF
--- a/lib/authentication.js
+++ b/lib/authentication.js
@@ -30,7 +30,7 @@ module.exports.loginRequired = function(req, res, next) {
     // Wipe req.user, just in case.
     req.user = undefined;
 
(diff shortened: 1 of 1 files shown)


rdegges:
Thank you =)

rdegges:
Cut a release with this included.

PR #66, opened 2015-04-20
add note about exports on windows

robertjd:


rdegges:
Thanks! I had no idea you could use the `set` command. Good find oO

robertjd:
Kudos to @adriatic, I'm just the messenger :)

adriatic:
I am used to get kudos for a lot more significant contributions - so I guess one gets them cheaper with age :-)

PR #76, opened 2015-07-07
[WIP] add optional validation callbacks for all fields

smartmike:
Just opening this as an quick concept, would love some feedback on it. Working on a project right now, we would want to validate various fields before registration. For this example, check the user's email matches a certain domain set in an environment variable.

The idea is a simple object matching the fieldnames which would validate true or false any way the user wants.

signup-validations.js:
```javascript
var addrs = require('email-addresses');

function checkEmailDomain(email) {
  var mail = addrs.parseOneAddress(email);
  return mail.domain === process.env.DOMAIN;
});

module.exports = {
  email: checkEmailDomain
};
```

server.js:
```javascript
var stormpath = require('express-stormpath');
var signupValidations = require('./signup-validations.js');

// ...

app.use(stormpath.init(app, {
  application: process.env.STORMPATH_APP_HREF, 
  fieldValidations: signupValidations
}));

```

Could also improve this to return a validation message rather than a generic one.

### DIFF
--- a/lib/controllers.js
+++ b/lib/controllers.js
@@ -50,6 +50,34 @@ function assertRequired(req, res, form) {
   };
 }
 
+/**
+ * @return {Function} Return a function which accepts an account hash and a
+ *   callback.
+ */
+function checkValidations(req, res, form) {
+  var view = req.app.get('stormpathRegistrationView');
+  var validations = req.app.get('stormpathFieldValidations');
+
+  return function(callback) {
+    async.each(Object.keys(FIELDS), function(key, next) {
+      if (validations[key] && typeof validations[key] === 'function') {
+        var valid = validations[key].call(null, form.data[key]);
+        if (!valid) {
+          helpers.render(view, res, { error: FIELDS[key] + ' invalid.', form: form });
+          req.app.get('stormpathLogger').info('User submitted a form with an invalid key: ' + FIELDS[key] + '.');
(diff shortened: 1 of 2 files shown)


robertjd:
This is a great idea!  Though I think it would be better as something like a "pre registration" hook, which would allow you to do asynchronous work (there may be other work that needs to be done in order to verify the attempted registration).

So the option might look more like this:

```javascript
app.use(stormpath.init(app, {
  application: process.env.STORMPATH_APP_HREF, 
  preRegistrationHandler: function(req,res,next,postedData){
    // postedData is a convenience object, so you don't have to 
    // did the properties off the req object.
    // call next when you are done, with null or an error
  }
}));

timothyej:
I will close this PR since it's getting old and is not included in our plans for this library. But we really appreciate your help and future contributions are very welcome!

smartmike:
No problem, sorry I didn't get further with it :)

timothyej:
Thanks :) Regarding the suggestion by @robertjd about a `preRegistrationHandler`, I've created an issue for it here: #299.

robertjd:
Thanks @smartmike for the original PR, and @timothyej for logging the issue.  @smartmike , we hope that the pre-registration handler will help you out.  We definitely want to give you control of the entire registration flow, and we think this is the best way to do that, without tying you to a specific validation implementation that our library would have to implement.  I hope all is well for you in express-stormpath land!

Issue #33, opened 2014-10-31
Password change page


lemieux:
As discussed in #20, express-stormpath should also provide a password change page. This should be almost the same as the password reset one, but instead of verifying a token, the current password should be provided. This page should only be accessible by authenticated users. I think it should also be possible to configure whether or not you should sign in again after the password change. Either way, all other sessions (or all sessions in case of a log out after the change)  should be nuked to prevent unauthorized access from another computer/browser.

We are most likely going to tackle this problem in our app in the next two weeks and since this is a generic thing, I think it should go in express-stormpath.

Thoughts?

denim2x:
I've come up with a possible fix; may I submit all code changes as a single commit or separately as multiple commits?

rdegges:
Hey @denim2x, it'd be great if you could submit a PR to the develop branch in one big PR. I'll review things.

Right now I'm still against providing this, as it doesn't make much sense to me.

Here's my full reasoning (wall of text follows):

For users who 'forget' their password, our existing password reset workflow is really the ideal way to allow a user to reset their password. We get the user to enter their username or email address, and then we send them an email with a link to click that takes them back to your site, and we provide a page there that allows them to reset their password.

This makes perfect sense.

This github issue is talking about providing a page so that users who are already logged in, and inside of a dashboard (presumably), can view a page to JUST change their own password to something else.

Usually something like this is 100% dependent on the application a user is building. For instance, with password reset stuff, it's easy to provide that as we're providing top-level URLs for an application, eg: /forgot, /reset, etc.

With a password change page, we'd have to start doing odd things that don't make sense for many applications (by default), eg: /users/password, or /accounts/password, or something similar.

This means we'd be expecting the developer to make a /users/ or /accounts page, and have an entire sort of user profile management type thing in their site.

Take github for instance, here's how the github 'password change page' looks:

<img width="1280" alt="screenshot 2015-07-02 18 46 17" src="https://cloud.githubusercontent.com/assets/90247/8490193/add55c1a-20ea-11e5-94d1-eb9f7c5a7076.png">

As you can see, there's a TON of context-specific information around this. Unlike a login page, registration page, or password reset page -- a password *change* page has TONS of specific site stuff, and doesn't make sense unless it's super duper customized.

Because of the above reasoning, and also because of how easy this is to implement on your own, I don't think it's really appropriate to have this sort of feature in this library.

If you wanted to make your own password change page, you could implement your own page (however you like) that just accepts a new password.

[...]

PR #77, opened 2015-07-07
Password change (#33)

denim2x:
Enables the user to change his/her password after logging in.


lhazlewood:
@robertjd how does this relate to the config/controller design we concluded a couple of weeks ago?

robertjd:
@lhazlewood we were defining the password reset workflow.  This appears to be a feature request for a change password page, for authenticated users.

PR #83, opened 2015-09-01
fix fields name in register form

suryod:
Small issue that causing the register form submit empty POST data.


rdegges:
Thanks!!

rdegges:
Sorry for the delay on this, by the way. With the new 2.x release, we've been busy cleaning up issues =)

PR #110, opened 2015-09-25
Fix auto expansion of custom data

erbridge:


rdegges:
I pushed this fix in manually, as we did some big refactoring which would have broken the merge! Thanks! Will cut a release tomorrow.

bassie1995:
This hasn't fixed the issue for me, it still gives me the Object with only an href in it.

rdegges:
This release hasn't been cut yet, sorry! Adding in testing around it.

bassie1995:
Thanks, will update when the release hits!

bassie1995:
Is this in 2.0.9? It still doesn't seem to work for me.

robertjd:
This *should* be fixed in 2.0.9, can you give me more context for your problem?  If you could share your code with me that would be really helpful.  If you don't want to share here you can contact us privately via support@stormpath.com

bassie1995:
Yep, my app.js is as follows: 
https://gist.github.com/bassie1995/9feafbc4fb72166c00f8.

Tried to use just expandCustomData: true as well.

I'll be glad to provide more info.

erbridge:
This is working for me now. I think you need:

```js
expand: {
  customData: true
}
```

not `expandCustomData: true`.

Also, you probably want to reset your API keys since they're completely exposed in that gist...

bassie1995:
Huh, that did fix it. I have tried that (or something close, I guess) before, but I'm glad it works now.

I'll be resetting the keys once we go live, as well :)

Issue #118, opened 2015-10-04
Issue with Social Login Google


dgisser:
When I try to login via google+, I get an "Uncaught TypeError: Cannot read property 'replace' of null".

In the function googleLogin() there is a line that says "var next = getQueryString('next').replace(/^\s+|\s+$/g, '');" Because this line only passes in one parameter to getQueryString, getQueryString always returns null and the .replace() function always runs on a null variable, which breaks the code. 

I will look into this more/create a pull request if I have time, but it seems that the solution would be to add the url before calling getQueryString.

I also feel like this googleLogin function is unnecessarily long.  The googleLogin function for flask with stormpath looks like 

  function googleLogin() {
    window.location.replace('https://accounts.google.com/o/oauth2/auth?response_type=code&client_id=/*client_id*/&scope=email+profile&include_granted_scopes=true&redirect_uri=/*redirect_uri*/');
  }

Why can't you make it just look like that?

rdegges:
Ok, merged this in. Also added you to the contributors page: https://github.com/stormpath/stormpath-express/commit/8b7eadb04507caf69e57d0b5cc46c97e5a8778f1

PR #119, opened 2015-10-04
fixed google plus integration

dgisser:
#118 This will fix issues with var next giving an error, especially since this variable is never used to begin with.

### DIFF
--- a/lib/views/google_login_form.jade
+++ b/lib/views/google_login_form.jade
@@ -6,19 +6,12 @@ script(type='text/javascript').
    * @param  {String} url   The URL to get the value from (optional)
    * @return {String}       The field value
    */
-  function getQueryString(field, url) {
-      var href = url ? url : window.location.href;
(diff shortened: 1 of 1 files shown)


rdegges:
Thanks! This makes sense. I think the original plan was to support redirection to the `?next=...` url with Google Login, but I don't think I found a way to do it using their API, so I left it half-implemented. Am happy to merge this in!

robertjd:
Perhaps we should implement an option in our config, such as `socialProviders.google.nextUri` ? This would be similar to how we handle all of our other redirects, and gives us a way to declare a post-google-login redirect (without relying on their API).

rdegges:
This has is specifically meant for people who want to hit a protected page, eg: `/secret`, then get redirected to login: `/login?next=%2Fsecret`, then log in with Google and get back to `/secret` automatically.

robertjd:
Ah, thanks for clarification.

PR #121, opened 2015-10-05
Social Login via API

### LINKED ISSUE #112: Social Login, via JSON API, for SPA clients
Right now the login controller looks for username and password, and then uses the OAuth authenticator to exchange those credentials for an access token.

We need to support social login from the Angular SDK, but our OAuth feature does not yet have a grant type for social providers (it does not have an authorization code grant flow).  Thus, for social login, we have to continue using a custom cookie-based session scheme.  We should re-use `helpers.createIdSiteSession()` for this use case.

The POST to `/login` from a SPA would look like this:

```
{
    providerData: {
        providerId: 'facebook',
        accessToken: 'from facebook'
    }
}
```

The controller will then make a call to `application.getAccount()` with this provider data.  You'll get an account object back, and pass it to `helpers.createIdSiteSession()` to create a session.

We will also need to expose, via a JSON GET api, the information that the SPA will need to do social login.  It needs to know which providers are available, and the client id's that will be needed for communicating with the provider.

That information is available in the `spConfig` object, as the underlying SDK pulls this information from the Stormpathpath application.  All we need to do is expose this so that the SPA can use it.

So the result of that GET may look like this:

```
  "socialProviders":{
    "google": {
      "clientId":""
    },
    "facebook": {
      "clientId":""
    },
    "linkedin": {
      "clientId":""
    }
  }
```

timothyej:
Fixes #112.

This PR adds support for social login from SPAs with for example the Angular SDK.

### Includes

* Support for login using `providerData` instead of username/password with the login API
* A new API route, `/oauth/providers`, that returns a list of all OAuth providers and their configurations
* Logouts with the `Accept` header set to `application/json` will respond with 200 instead of a redirect

### Notes

* Is the route `/oauth/providers` the best route for this purpose?

### How to verify

1. Setup a server with this branch
2. Verify the login, logout, etc. works with the AngularJS SDK: <https://github.com/stormpath/stormpath-sdk-angularjs/pull/48>
3. Run tests


rdegges:
Looking good so far =)

robertjd:
I've reviewed the code changes and I like it, I don't have any comments on the code.  But please add the following to this PR:

* JSDoc comments for new functions

* Tests for these cases:

 * That the `/oauth/providers` endpoint is serving the expected object, if the application is mapped to a directory that has provider data

 * That the `/oauth/providers` endpoint returns an empty object, `{}`, when there are no account stores with provider data on the application

timothyej:
Thanks @robertjd, I will fix the docs and tests then this should be ready for final review :)

timothyej:
@rdegges @robertjd @typerandom This is ready for final review! I've added some jsdocs and some tests and fixed the conflicts. I don't know why the Travis build fails though?

lib/controllers/login.js
```diff
@@ -6,7 +6,8 @@ var uuid = require('uuid');
 
 var forms = require('../forms');
 var helpers = require('../helpers');
-var oauth = require('../oauth');
+var oauth = require('../oauth')
```
typerandom: Missing `;`.

lib/controllers/login.js
```diff
@@ -50,6 +51,19 @@ module.exports = function(req, res) {
   res.locals.status = req.query.status;
 
   if (req.method === 'POST' && accepts === 'json') {
+    if (req.body.providerData) {
```
typerandom: How about adding `var providerData = req.body.providerData;` then `if (providerData) {`, etc, and using that instead of repeating it?

lib/controllers/logout.js
```diff
@@ -14,11 +14,17 @@ var middleware = require('../middleware');
  */
 module.exports = function(req, res) {
   var config = req.app.get('stormpathConfig');
+  var accepts = req.accepts([ 'json' ]);
```
typerandom: That array style looks very inconsistent. Haven't seen it before.

test/controllers/test-logout.js
```diff
@@ -52,4 +52,26 @@ describe('logout', function() {
       .expect('Set-Cookie', /refresh_token=;/)
       .end(done);
   });
+
+  describe('when Accept header is set to text/html', function() {
```
typerandom: Looks inconsistent to use mixed case for describe/it statements.

test/controllers/test-oauth-providers.js
```diff
@@ -0,0 +1,151 @@
...
+
+  after(function(done) {
+    helpers.destroyApplication(stormpathApplication, done);
+  });
+
+  describe('when no social directories exist', function() {
+    before(performRequest);
+
+    it('should return an object', function() {
+      assert.equal(typeof response === 'object', true);
```
typerandom: Clearer to use `assert.typeOf`.

test/controllers/test-oauth-providers.js
```diff
@@ -0,0 +1,151 @@
...
+    });
+
+    it('should have "facebook.enabled" set to false', function() {
+      assert.equal(response.hasOwnProperty('facebook'), true);
```
typerandom: Clearer to use `assert.isTrue`.

test/controllers/test-oauth-providers.js
```diff
@@ -0,0 +1,151 @@
...
+      assert.equal(response.facebook.enabled, false);
```
typerandom: Clearer to use `assert.isFalse`.

test/controllers/test-oauth-providers.js
timothyej: Nice, didn't know about that one :+1:

test/controllers/test-oauth-providers.js
timothyej: Btw, which assertion library are we using? Because I can't find these functions here: <https://nodejs.org/api/assert.html> :(

typerandom: Chai.

http://chaijs.com/api/assert/

test/controllers/test-logout.js
timothyej: Why, just because it starts with lowercase doesn't mean all letters have to be lowercase? The accept header is written with a capital `A` so it only makes sense in my opinion.

test/controllers/test-oauth-providers.js
timothyej: Oh, that wasn't very clear 😂

test/controllers/test-logout.js
typerandom: You're right. My bad. I was just used to lowercase being used as the general style, but I can see now that it's being used differently.

lib/controllers/login.js
timothyej: Good catch :+1:

test/controllers/test-oauth-providers.js
typerandom: Actually, when looking at this now it seems like in `stormpath-express` Node.js built in assertion library isn't being used. We only use Chai in the SDK. Which is very inconsistent.

lib/controllers/login.js
timothyej: Much better :)

lib/controllers/logout.js
timothyej: `login.js` uses this style, but I will change them both to not use spaces. I also decided to add `html` to the list to be consistent and be clear that this route supports both html and json.

timothyej:
Thanks for the review, @typerandom! I've addressed your comments in my latest commit.

robertjd:
Closing in favor of #171

Issue #106, opened 2015-09-24
Login view config doesn't respect enabled


erbridge:
I'm trying to implement my own login view and switching `web.login.enabled = false` doesn't disable the login endpoint.

```js
{
  expand: {
    customData: true,
  },

  website: true,

  web: {
    register: {
      enabled:       true,
      autoAuthorize: true,
      fieldOrder: [
        'username',
        'email',
        'givenName',
        'surname',
        'password',
        'passwordConfirm',
      ],
    },
    login: {
      enabled: false,
    },
    logout: {
      enabled: true,
    },
    forgotPassword: {
      enabled: true,
    },
    changePassword: {
      enabled:   true,
      autoLogin: true,
    },
  },
}
```

typerandom:
@erbridge Thank you for reporting this issue.

The issue here is that `website: true` conflicts with the `web` configuration.

What `website: true` actually does it that it enables the `web` endpoints `register`, `login`, `logout` and `me`. What you can do meanwhile is remove `website: true` and add `register` and `me` to your list of enabled `web` endpoints.

I.e. so it would look like this:

```js
{
    expand: {
        customData: true
    },
    web: {
        register: {
            enabled: true,
            autoAuthorize: true,
            fieldOrder: [
                'username',
                'email',
                'givenName',
                'surname',
                'password',
                'passwordConfirm'
            ]
        },
        login: {
            enabled: false
        },
        logout: {
            enabled: true
        },
        forgotPassword: {
            enabled: true
        },
        changePassword: {
            enabled: true,
            autoLogin: true
        },
        me: {
            enabled: true
        },
        register: {
            enabled: true
        },
    }
}
```

Let me know if this helped you.

typerandom:
There's now a PR in our config module that contains a fix for this issue. Once we have merged that in and have updated the integration to use that version, this issue will be resolved.

robertjd:
This should be fixed in 2.1.0, please re-open if that is not the case.

PR #122, opened 2015-10-06
Refactor: Configuration

typerandom:
Refactors the way that configuration is handled so that configuration that was previously in the SDK is now in the integration.

**Important:** This may only be merged once there is a new release for [the stormpath module](https://github.com/stormpath/stormpath-sdk-node) that contains the changes from PR https://github.com/stormpath/stormpath-sdk-node/pull/213.

#### What this fixes

* Integration config now in integration instead of SDK.
* Social providers now enriched automatically instead of relying on hard-coded defaults in config. Because of this we now have to check if the provider exists before using it. See: https://github.com/stormpath/express-stormpath/compare/refactor-client-config?expand=1#diff-ef4e9b51bbe32bc78fc66cfd8d410182R127.

(additional)

* SDK errors are now emitted through express app channel `stormpath.error`.

#### How to verify

1. Run `$ npm test`.
2. Either create a new, or use an existing `express-stormpath` app.
3. Assert that both configuration options `website` and `api` enable the correct endpoints for those options.
4. Assert that social providers are being populated by social provider directories mapped to your application.


robertjd:
I was reviewing this PR, and I'm happy with what I'm seeing.  I updated the Node SDK dependency and merged in master.  The get-user test is going to need to be fixed, it is depending on the old way of loading a default configuration

robertjd:
There is one dilemma I’d like to point out.  With this PR, we don’t bind express routes until after the configuration is resolved, which is async.  This is annoying if you’re using a live-reload environment, because a route will return 404 before the configuration is loaded, but then return our framework response once the configuration is loaded.

We should decide if we care about this use-case.  But I’ll stay that as developer who supports our Angular SDK (and the related example application) I’m frustrated by this break in my live-reload workflow.

If we were to support this use case, we’d have to synchronously load the basic configuration that tells us which routes we want to bind to.  We would then eagerly bind to the routes, then assume that any calls to the underlying SDK will be deferred until the client is actually ready (the config has been resolved).  This appears to be the case with the 0.13.0 release of the Node SDK.

typerandom:
@robertjd Could you explain what you mean with that the "get-user test needs to be fixed"?

Also, when it comes to live-reload, what we could do is do so that the middleware blocks until the SDK is ready. I.e. we inject the middleware, but does not continue until the middleware is ready.

I'll start working on a fix for this and let you know when you can review it :)

robertjd:
> @robertjd Could you explain what you mean with that the "get-user test needs to be fixed"?

These tests are failing right now, because they're trying to require the old config.json:

https://github.com/stormpath/express-stormpath/blob/master/test/helpers/test-get-user.js

I love fixing tests, so I will fix those.

> Also, when it comes to live-reload, what we could do is do so that the middleware waits until the SDK is ready. I.e. we inject the middleware, but doesn't let the middleware continue until the SDK is ready.

That is what I was thinking.  Do you have an idea of how you'll go about it?  It seems that we'd have to do some synchronous loading of any provided configuration files.

typerandom:
>These tests are failing right now, because they're trying to require the old config.json. I love fixing tests, so I will fix those.

I hate fixing tests:) But awesome, I'll do my things and if you get time to fix them then please do. Else I'll just do it:)

>That is what I was thinking. Do you have an idea of how you'll go about it? It seems that we'd have to do some synchronous loading of any provided configuration files.

We can still do everything async. All we need to do is that if the SDK isn't ready, then await the `ready` event before calling next on our middleware. I.e. so that all requests that use our middleware will queue until we get our `ready` event.

robertjd:
I want to review this but won't have time tonight, will pickup tomorrow.

typerandom:
I have now updated so that the new Stormpath module v0.13.2 is in use. Just noticed that I had a typo in my commit message saying version 0.13.1, but it should be 0.13.2 :)

robertjd:
I like the YAML file, that is a lot easier to read than JSON.. :)

There is a failing case for the unerlying option parser, can you take a look at this case I wrote?

https://github.com/stormpath/stormpath-node-config/commit/a955f6bd9670c8a0c338ca6df7875558ec9f0862

Thanks!

typerandom:
Yeah, I also like it a lot more :) Ah, this is related to this issue: https://github.com/stormpath/express-stormpath/issues/106

Right?

robertjd:
Yes!  I was trying to find that.. but my mind knew it existed, hence the test :)

typerandom:
Awesome! I have an idea on how to fix that, so will look into it :)

robertjd:
I merged https://github.com/stormpath/stormpath-node-config/pull/3 - do we need to do much here?

package.json
```diff
@@ -28,13 +28,13 @@
     "csurf": "^1.8.0",
     "deep-extend": "^0.4.0",
     "express": "^4.10.6",
-    "express-stormpath-config": "0.0.2",
```
robertjd: Do we delete this package after this PR is merged?

test/controllers/test-email-verification.js
```diff
@@ -38,20 +38,32 @@ describe('email verification', function() {
...
+        stormpathApplication = app;
+
+        helpers.setEmailVerificationStatus(stormpathApplication, 'ENABLED', done);
+      });
     });
   });
 
   after(function(done) {
-    helpers.destroyApplication(stormpathApplication, done);
+    if (stormpathApplication) {
```
robertjd: Why this if clause?

robertjd:
@typerandom please merge in master to handle conflicts, then ping me for one last review before we merge this

package.json
typerandom: Yes, we can do that.

test/controllers/test-email-verification.js
typerandom: I can't remember exactly. I'm just going to remove it:)

typerandom:
@robertjd It's merged in:) Please review.

robertjd:
So close!  I'm signing off on the code quality (disclaimer: I'm ignoring all the declared but un-used variables, but only because we haven't yet merged our eslint config :) )

But now I've got an error, @typerandom can you help me reproduce?   I'm getting an error from the extend library, using my demo angular application.

Please clone this repo and enter this folder and follow the instructions to setup this demo application:

https://github.com/stormpath/stormpath-sdk-angularjs/tree/master/example/dashboard-app

Then use `npm link express-stormpath` to link that demo to this branch.

When I run `grunt serve`, I'm getting this error:

```
/Users/robert/projects/express-stormpath/node_modules/stormpath-config/node_modules/cloneextend/index.js:48
 b_pos=context.indexOf(b);
               ^
TypeError: undefined is not a function
    at extend (/Users/robert/projects/express-stormpath/node_modules/stormpath-config/node_modules/cloneextend/index.js:48:16)
    at /Users/robert/projects/express-stormpath/node_modules/stormpath-config/lib/strategy/EnrichIntegrationFromRemoteConfigStrategy.js:83:15
    at Object.wrapGetResourceResponse [as _wrapGetResourceResponse] (/Users/robert/projects/express-stormpath/node_modules/stormpath/lib/ds/DataStore.js:89:12)
    at onGetResourceRequestResult (/Users/robert/projects/express-stormpath/node_modules/stormpath/lib/ds/DataStore.js:278:13)
    at Request.onRequestResult [as _callback] (/Users/robert/projects/express-stormpath/node_modules/stormpath/lib/ds/RequestExecutor.js:110:7)
    at Request.self.callback (/Users/robert/projects/express-stormpath/node_modules/stormpath/node_modules/request/request.js:123:22)
    at Request.emit (events.js:110:17)
    at Request.<anonymous> (/Users/robert/projects/express-stormpath/node_modules/stormpath/node_modules/request/request.js:1047:14)
    at Request.emit (events.js:129:20)
    at IncomingMessage.<anonymous> (/Users/robert/projects/express-stormpath/node_modules/stormpath/node_modules/request/request.js:998:12)
```

typerandom:
@robertjd I'm looking into this right now.

typerandom:
@robertjd This issue is now fixed:)

Btw, the un-used variables, I guess you referring to things in tests? Because everywhere else we should use strict mode and and I haven't seen those warnings. Or how are you seeing these, did you run eslint over or are you using something else?

robertjd:
Awesome, thanks!  We can merge this PR after we fix this issue:

https://github.com/stormpath/stormpath-sdk-node/pull/253

Regarding un-used vars, there are some in lib/stormpath.js.  I've taken the .eslintrc from the eslint branch and put it in my home folder, so it's being used by default now in my sublime :)

robertjd:
@typerandom woot!

PR #137, opened 2015-10-09
Fixes #136 - Google login does not create session

### LINKED ISSUE #136: Google Provider does not create site session on login
The facebook and linkedin providers create a site session cookie on login, but the google provider does not.  This results in a user getting a successful login callback from google, but not be logged in.

ldhertert:


robertjd:
Thanks @ldhertert !  I opened & closed via another PR within this repo so that Travis could run the tests (we use secret keys in Travis which don't work with fork'd PR's, super annoying)

erbridge:
You could disable the tests that need the keys: http://docs.travis-ci.com/user/pull-requests/#Security-Restrictions-when-testing-Pull-Requests

robertjd:
All of our tests need keys :) we've purposely decided to use the real API (not a mock) in our tests.

erbridge:
Then disable travis for pull requests, perhaps (a toggle in the project settings on travis-ci.org)? Ignoring big red crosses seems like a bad practice to me.

robertjd:
The second PR strategy is a way of *not* ignoring the big red cross :) 

Notice how the second PR has caused this PR to have a nice green checkmark, because it's the same test run against the same commit.

From a workflow perspective, we really like issuing pull requests and having them run tests.  Unfortunately this breaks down with pull requests from forks.  I'm open to ideas.

PR #147, opened 2015-10-12
Making api-authentication-required expose apiKey into req and res

pierredavidbelanger:
A very simple PR that makes api-authentication-required expose the successfully authenticated ApiKey into the req and res.

### DIFF
--- a/lib/middleware/api-authentication-required.js
+++ b/lib/middleware/api-authentication-required.js
@@ -19,8 +19,10 @@ module.exports = function(req, res, next) {
   var logger = req.app.get('stormpathLogger');
 
   // Wipe user data in case it was previously set by helpers.getUser().
+  req.apiKey = undefined;
   req.user = undefined;
   req.permissions = undefined;
+  res.locals.apiKey = undefined;
   res.locals.user = undefined;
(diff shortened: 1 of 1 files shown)


lib/middleware/api-authentication-required.js
```diff
@@ -42,8 +44,10 @@ module.exports = function(req, res, next) {
           return res.status(401).json({ error: 'Invalid API credentials.' });
         }
 
+        res.locals.apiKey = authResult;
```
robertjd: This should be `res.locals.authenticationResult`, as that is the type of object that is being assigned.

pierredavidbelanger: Indeed, the type of `authResult` is `AuthenticationResult`, but its `href` is something like:
`'https://api.stormpath.com/v1/apiKeys/ABCDEF123456789'`, so I was confused when naming the property.

I will do as you said. You can close this PR, I will issue a fresh new one on top of the master branch.

robertjd:
Thanks @pierredavidbelanger - I'm going to close this one, in favour of #148.  In the future you can push commits to the same branch/PR - github will figure out how to display things nicely :)

PR #149, opened 2015-10-13
Fixing JSON error responses in registration controller

robertjd:
Pass through the error from the underlying SDK

### DIFF
--- a/lib/controllers/register.js
+++ b/lib/controllers/register.js
@@ -89,6 +89,20 @@ function applyDefaultAccountFields(config,req) {
     req.body.surname = 'Anonymous';
   }
 }
+/**
+ * Use this method to render JSON error responses, for errors that arise from
+ * the underlying Node SDK
+ *
+ * @function
+ *
+ * @param {Object} err - The error from the Node SDK
+ * @param {Object} res - The http response
+ */
+function defaultJsonErrorResponder(err,res) {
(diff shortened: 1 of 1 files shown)


lib/controllers/register.js
```diff
@@ -89,6 +89,20 @@ function applyDefaultAccountFields(config,req) {
     req.body.surname = 'Anonymous';
   }
 }
+/**
```
timothyej: How about moving this into its own file? :)

robertjd: I was thinking about that.  My general approach is to not move things into files until I need to require the function by multiple files.  But now I'm wondering if having it in a separate file would encourage others to notice that it's there and to start using it?  Because I do want us to get a control on error handling in this library..

Where would you put it? I image putting it in the `helpers/` folder.

rdegges: FWIW: Initially I imagined us having a `account-defaults.json` file, or something similar, which basically gets extended with `req.body` data to supplement in any missing (but optional) fields.

timothyej: The `loginResponder` is in `helpers/` so I think that's a good place for this :)

robertjd: After looking at this again, the fallback error message is specific to this registration behavior.  I don't see a need to break this out to a file (yet).

PR #154, opened 2015-10-15
Fix tests that aren’t cleaning up application resources

robertjd:
Most of these were my fault, oopsies :)

I also added the package name and a per-process uuid to the application name, so that (a) I could see how many apps were being leftover per run and (b) so that others can know that this is our library


robertjd:
Spoke with @typerandom during standup today and he said this looked good, merging.  Ping @dogeared

PR #159, opened 2015-10-17
post*Handlers are not invoked for google (and likely for linkedin)

cdaniel:
handlers are retrieved from the application, but there is no trace of them being ever set Let's get them from the config.


rdegges:
Thanks for the PR! Looks great, merging in now.

PR #174, opened 2015-10-29
Feature: OAuth Password grant and refresh grant flow for OAuth endpoint

### LINKED ISSUE #162: Implement password grant & refresh grant flow for oauth endpoint
This library needs to support `grant_type=password` on the `/oauth/token` endpoint.  The implementation of this handler will look very similar to this:

https://github.com/stormpath/express-stormpath/blob/master/lib/controllers/login.js#L53-L56

But rather than settings cookies we need to respond with the OAuth-specificed response body, which is a JSON object that looks like this:

```
{
  "access_token": "eyJraWQ...",
  "expires_in": 1800
  "refresh_token": "eyJraWQiOiI...Y",
  "token_type": "Bearer",
}
```

As such we also need to implement the `refresh_token` grant flow, allowing the user to exchange their refresh token for an access token

The `apiAuthenticationRequired` middleware will also need to be modified to accept these access tokens via `Authorization: Bearer <access_token>`

typerandom:
Adds support for the OAuth password grant and refresh grant flow for the OAuth endpoint.

Fixes #162.


test/middlewares/test-api-authentication-required.js
```diff
@@ -101,9 +100,46 @@ describe('apiAuthenticationRequired', function() {
           .set('Authorization','Bearer ' + access_token )
           .expect(200,successResponse,done);
       });
+    });
+  });
 
+  it('should return error response if invalid oauth credentials are supplied', function(done) {
```
timothyej: What about wrapping this in a `describe` like this:

```javascript
describe('when invalid oauth credentials are supplied', function() {
  it('should return an error response');
});
```

test/middlewares/test-api-authentication-required.js
```diff
@@ -101,9 +100,46 @@ describe('apiAuthenticationRequired', function() {
...
+    var app = helpers.createStormpathExpressApp({
```
timothyej: The setup of the Express app is redundant for each test case. Maybe extract it into a `beforeEach` function?

lib/middleware/authentication-required.js
```diff
@@ -18,7 +18,7 @@ var deleteCookies = require('./delete-cookies');
  */
 module.exports = function(req, res, next) {
   var logger = req.app.get('stormpathLogger');
-
+
```
timothyej: Why did you add two spaces on this empty line? 😂

robertjd:
I updated this PR with a merge from master, and eslint fixes.  But as I review this I find myself wanting to clean up a lot of our OAuth code, as there are some weird edge cases here (like being able to use Basic Auth to authenticate through the oauth endpoint, but not actually get a token back).  It also appears that the Node SDK isn't returning the correct error code text for the OAuth specific errors.

I'll revist next week, so please consider this PR a work in progress for the moment

robertjd:
I've added some work to this PR.  Here is what's new:

* Tests have been significantly changed for asserting our Oauth2 features.  They weren't actually asserting that the password grant flow or refresh grant flow was working.  We also didn't have a correct test to assert the client_credentials workflow was working, which it was not in this PR.

* This PR now depends on [stormpath-sdk-node/oauth-fixes](https://github.com/stormpath/stormpath-sdk-node/tree/oauth-fixes) for a few small changes to how we handle Oauth errors during credential exchange.  @typerandom if you could please review [#270](https://github.com/stormpath/stormpath-sdk-node/pull/270) that would be awesome.  We will need to relase 0.13.5 of the Node SDK and then update package.json in this PR accordingly.

* I updated the Authentication section of this documentation to better describe all the Authentication flows including this new Password grant flow.  Ping @rdegges as I'd love to get your feedback.

@typerandom I'd like it if you could review my new work on this PR as well.

Thanks to everyone for helping to get this one out the door!  This a big step towards getting *all the things!* into this library.

wimolivier:
Hi Robert,

Any idea when the full password grant flow will be complete?

Thanks,
Wim

docs/authentication.rst
```diff
@@ -3,18 +3,30 @@
 Authentication
 ==============
 
+This library offers several options for securing your server and authenticating
```
rdegges: Instead of saying 'server' (that makes me think of a hardware box), we should probably say 'service'.

docs/authentication.rst
```diff
@@ -24,21 +36,13 @@ required, or continue into your middleware::
...
-
-    client.getApplication('your app href',function (err,application){
-        application.getOAuthPolicy(function (err,policy){
-            policy.accessTokenTtl = "PT1D"; // one day
-            policy.save();
-        });
-    });
+If you need to change the expiration time of the cookies, please login to the
+Stormpath Admin Console and find the OAuth policy inside of your Stormpath
+Application.  Then change the expiration time of the Refresh Token.
```
rdegges: Why are we mentioning the refresh token expiration time here? We should explain it first, I think =)

docs/authentication.rst
```diff
@@ -49,11 +53,17 @@ To change these settings, you should invoke a node client directly::
     This works by utilizing the Express `router`_.
 
 
-API Authentication: Basic Auth
-------------------------------
+Issuing API Keys
+----------------
+
+If you are building an API service you will need to distribute API keys to your
```
rdegges: Put a comma after 'service'

docs/authentication.rst
```diff
@@ -49,11 +53,17 @@ To change these settings, you should invoke a node client directly::
...
+the next sections, but we need to provision API keys for your developers first.
 
-For any account in your application, you can provision API Keys for those accounts.
-Here is an example::
+While your service may be an API service, you will still need to provide a
+basic website that developers can use to obtain their keys.  Here is an example
+of how you can create an API Key for the currently logged in user::
 
     app.post('/apiKeys', stormpath.loginRequired, function (req, res) {
       req.user.createApiKey(function (err,apiKey){
```
rdegges: Space in the code sample after the `err` argument.

docs/authentication.rst
```diff
@@ -65,8 +75,45 @@ Here is an example::
       });
     });
 
-Your API clients can then use this key to authenticate against your API.  This
-is how you would protect the secret endpoint with basic authentication::
+This is a naive example which simply prints out the API Keys for them, but
```
rdegges: 'prints out the API Key for the user...'

docs/authentication.rst
```diff
@@ -65,8 +75,45 @@ Here is an example::
...
+once they have the keys they will be able to authenticate with your API.
+
+For more information on API Keys, please see
+`Using Stormpath for API Authentication`_
+
+Enabling API Authentication
+---------------------------
+
+Regardless of the authentication strategy that you choose, you will need to
```
rdegges: Mmmm. This isn't true. Basic Auth will work without this option being turned on. This only hooks up the oauth2 routes.

docs/authentication.rst
```diff
@@ -65,8 +75,45 @@ Here is an example::
...
+
+HTTP Basic Authentication
+-------------------------
+
+This strategy makes sense if you are building a simple API service that does
+not have complex needs around authorization and resource control.  This strategy
+is simple because the developer simply supplies their API keys on every request
+to your server.
+
+Once the developer has the API keys, they will use them to authenticate with your
```
rdegges: has *their* API key

docs/authentication.rst
```diff
@@ -74,89 +121,89 @@ is how you would protect the secret endpoint with basic authentication::
...
 
-    Authentication: Bearer <Base64UrlSafe(apiKeyId:apiKeySecret)>
+OAuth2 Client Credentials
+-------------------------
 
-You can use Curl to achieve the same request::
+If you are building an API service and you have complex needs around
+authorization and security, this strategy should be used.  In this situation
+the developer does a one-time exchange of their API Keys for an access token.
+This access token is time limited and must be periodically renewed.  This adds a
```
rdegges: periodically *refreshed*

docs/authentication.rst
```diff
@@ -74,89 +121,89 @@ is how you would protect the secret endpoint with basic authentication::
...
+layer of security, but at the cost of being less simple than HTTP Basic
```
rdegges: remove the word ', but'

docs/authentication.rst
```diff
@@ -74,89 +121,89 @@ is how you would protect the secret endpoint with basic authentication::
...
-  via subsequent requests.
-- ``token_type`` - This will always be ``'bearer'``.
+  on future requests.
+- ``token_type`` - This will always be ``"bearer"``.
 - ``expires_in`` - This is the amount of seconds (*as an integer*) for which
   this token is valid.
 
-With that token, we can make a request of our secret endpoint and supply the
-token as the bearer header in the request::
+With that token you can now make requests of your API.  This request is simpler,
```
rdegges: *this token*

rdegges: *to your API*

docs/authentication.rst
```diff
@@ -74,89 +121,89 @@ is how you would protect the secret endpoint with basic authentication::
...
+the only thing you need to do is supply the token in the ``Authorization`` header
```
rdegges: *as the only thing*

docs/authentication.rst
```diff
@@ -74,89 +121,89 @@ is how you would protect the secret endpoint with basic authentication::
...
+      res.json({
+        message: "Hello, " + req.user.fullname
+      });
+    });
 
-You can control the expiration time of the tokens that are created for the
-client credentials workflow.
+By default the access tokens are valid for one hour.  If you want to change
+the expiration of these tokens you will need to configure it in the server
+configuration, like this::
```
rdegges: We do this via the oauth policy on the app now, yea?

docs/authentication.rst
```diff
@@ -164,4 +211,66 @@ Here is an example::
     }));
 
 
+OAuth2 Password Grant
+---------------------
+
+This is the authentication strategy that you will want to use for mobile clients.
+In this situation the end-user supplies their username and password to your
+mobile application.  The mobile application sends that username and password to
+your server, which then verifies the password with Stormpath.
```
rdegges: We should not use 'server' here, I think.

docs/authentication.rst
robertjd: Only for the password-based flow. The client-credentials flows is still done in our SDK.  We haven't implemented client credentials in the core API yet.

PR #186, opened 2015-11-03
Fix: Social providers UI box begin shown when no providers are present

### LINKED ISSUE #182: Fix UI Bug when no social providers are present
ui bug no social providers express-stormpath v2.1.0

typerandom:
Fixes so that the social providers UI box is only shown if there are social providers present and enabled.

Fixes #182.

### DIFF
--- a/lib/controllers/login.js
+++ b/lib/controllers/login.js
@@ -1,5 +1,6 @@
 'use strict';
 
+var _ = require('lodash');
 var extend = require('deep-extend');
 var stormpath = require('stormpath');
 var uuid = require('uuid');
@@ -33,10 +34,15 @@ module.exports = function(req, res) {
       options = {};
     }
 
+    var hasSocialProviders = _.any(config.socialProviders, function (socialProvider) {
+      return socialProvider.enabled;
+    });
+
     extend(options, {
       form: form,
       formActionUri: formActionUri,
-      oauthStateToken: oauthStateToken
+      oauthStateToken: oauthStateToken,
+      hasSocialProviders: hasSocialProviders
     });
 
     helpers.render(req, res, view, options);
--- a/lib/views/login.jade
+++ b/lib/views/login.jade
@@ -48,7 +48,7 @@ block body
                     receive an email shortly.
           br
         .box.row
-          if socialProviders
+          if hasSocialProviders
             - var cls = 'small col-sm-8'
           else
             - var cls = 'large col-sm-12'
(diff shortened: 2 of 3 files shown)


lib/controllers/login.js
```diff
@@ -33,10 +33,24 @@ module.exports = function(req, res) {
...
+    if (config.socialProviders) {
+      for (var name in config.socialProviders) {
+        var provider = config.socialProviders[name];
+        if (provider.enabled) {
+          hasSocialProviders = true;
+          break;
+        }
+      }
+    }
+
```
robertjd: I like to do this with filter:

```
var hasSocialProviders = Object.keys(config.socialProviders).filter(function(name){
  return config.socialProvider[name].enabled;
}).length > 0
```

I'm not attached to either way, just sharing.   Underscore also has some nice functions for making these one-liner operations.

typerandom: Thanks! I definitely agree that there are definitely nicer solutions:) I'm not that much for iterating over an object using Object.keys like that. But if you're familiar with lodash they have pretty nice helper methods for collections such as this. I.e:

```js
var hasSocialProviders = _.any(config.socialProviders, function (socialProvider) {
  return socialProvider.enabled;
});
```

I'll let you know when I have updated this.

timothyej: I agree with @robertjd, and I don't think that we should include underscore/lodash for this.

typerandom: I totally disagree with you. It's just such a big difference in readability. Also underscore is apparently already used in the project and has the method [any()](http://underscorejs.org/#some) that works in exactly the same way as `lodash.any()`.

timothyej: Nope, [`some()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) is a native method on the `Array` prototype and you don't need either underscore or lodash for that. Besides, what I can see none of them are included in this project and including them just for this would be overkill in my opinion.

typerandom: Yes, for an array. Which this is not.

Also, I thought we already had `underscore` included, but just checked and we don't. So I'll be adding `lodash` instead. I don't think including it is overkill in any sense, because in this case it increases readability and it's very likely that it will be able to improve a lot of the other code too. Though, if it was a module that only did one thing, then I'd definitely agree with you. But in this case I'd say it adds a lot of value to have included in the project.

Also, in the `stormpath` module we're currently using both `underscore` and `lodash`.

timothyej: Yeah, you're right!

However, I just think that we should try to keep the dependencies down as it adds a lot of maintenance work to keep them updated.

typerandom: I agree with that. If they are useless dependencies. But for those that add value, I'd say it's worth it.

robertjd: I like lodash :)

PR #191, opened 2015-11-05
Feature: Comply with framework registration spec

### LINKED ISSUE #183: Remove Fields by Default
If you have the `website: true` option enabled, and nothing else, the registration page looks like this:

screenshot 2015-11-02 10 36 08

As you can see, there are lots of fields there. Previously, we only showed the:

- first name
- last name
- email
- password

Fields, which is what we should return to doing. This way users (by default) aren't overwhelmed with all of the fields available.

typerandom:
Complies with the [framework spec PR](https://github.com/stormpath/stormpath-framework-spec/pull/15) by removing certain registration fields from the config file and patching places where these fields were used.

Closes #183

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -6,10 +6,29 @@ Change Log
 
 All library changes, in descending order.
 
-Version 2.2.1
+Version 2.3.0
 --------------
 
 **Not Yet Released**
+
+May fixes for the registration field configuration:
+
+- Custom fields, as defined in the ``register`` block, will now appear in our
+  default registration form (they were not appearing before).
+
+- Custom fields now *must* be defined in the ``register`` block, otherwise the
+  data will be rejected during account creation.
+
+- Only the First Name, Last Name, Email, and Password fields are shown by
+  default (the middle name and username fields are no longer shown be default).
+
+- Added an ``enabled`` property to all fields, allowing you to selectively
+  disable any of the default fields.
+
+- The default value for first name and last name is now "UNKNOWN", if not
+  provided and not required during registration.
+
+Please see the Registration section of this documentation for more information.
 
 
 Version 2.2.0
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -11,7 +11,7 @@ Version 2.3.0
 
 **Not Yet Released**
 
-May fixes for the registration field configuration:
+Many fixes for the registration field configuration:
 
 - Custom fields, as defined in the ``register`` block, will now appear in our
   default registration form (they were not appearing before).
@@ -20,7 +20,7 @@ May fixes for the registration field configuration:
   data will be rejected during account creation.
 
 - Only the First Name, Last Name, Email, and Password fields are shown by
-  default (the middle name and username fields are no longer shown be default).
+  default (the middle name and username fields are no longer shown by default).
 
 - Added an ``enabled`` property to all fields, allowing you to selectively
   disable any of the default fields.
--- a/docs/registration.rst
+++ b/docs/registration.rst
@@ -26,7 +26,7 @@ we will cover them in detail below:
     {
       web: {
         register: {
-          enable: true,   // Explicit enable, if not using { website: true }
+          enabled: true,   // Explicit enable, if not using { website: true }
           uri: '/signup',  // Use a different URL
           nextUri: '/',    // Where to send the user to, if auto login is enabled
           fields: {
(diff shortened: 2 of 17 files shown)


lib/controllers/register.js
```diff
@@ -81,11 +81,13 @@ function defaultJsonResponse(req, res) {
  */
 
 function applyDefaultAccountFields(config, req) {
-  if (!config.web.register.fields.givenName.enabled && !req.body.givenName) {
+  var registerFields = config.web.register.fields;
+
+  if ((registerFields.givenName && !registerFields.givenName.enabled) && !req.body.givenName) {
```
timothyej: Why the extra parentheses here? :)

lib/controllers/register.js
```diff
@@ -81,11 +81,13 @@ function defaultJsonResponse(req, res) {
...
     req.body.givenName = 'Anonymous';
   }
 
-  if (!config.web.register.fields.surname.enabled && !req.body.surname) {
+  if ((registerFields.surname && !registerFields.surname.enabled) && !req.body.surname) {
```
timothyej: And why extra parentheses here too?

lib/controllers/register.js
typerandom: Fixed:)

robertjd:
This introduces an `enabled` property, which is not something that we have in the spec for registration fields.  My understanding is that a field is either an object or null.  

If it is null, it is not shown.

If it is an object, we show the field.

If the object has the `required` property, we use form validation to force the user to provide the value.

For the `givenName` and `surname` fileds, if `required` is `false`, we populate the value 'UNKNOWN' before we submit the account data to the API (because the API requires it).  Note: we are currently using 'Anonymous', but I do believe 'UNKNOWN` is better and this is what we've defined in the framework spec.

Ping @rdegges on all this ^

rdegges:
Agree w/ Robert -- this is also the old functionality.

robertjd:
Update: after discussing it further, we decided to go the route of using the `enabled` property because trying to extend layers of configuration, where you can define a field as null, would be madness.  Also, with the enabled strategy we can define all of our default supported fields in our config.yaml, but just disable the ones that we don't want to render by default.

This PR has been updated accordingly.

docs/changelog.rst
```diff
@@ -6,11 +6,30 @@ Change Log
 
 All library changes, in descending order.
 
-Version 2.2.1
+Version 2.3.0
 --------------
 
 **Not Yet Released**
 
+May fixes for the registration field configuration:
```
rdegges: *Many

docs/changelog.rst
```diff
@@ -6,11 +6,30 @@ Change Log
...
+
+- Custom fields, as defined in the ``register`` block, will now appear in our
+  default registration form (they were not appearing before).
+
+- Custom fields now *must* be defined in the ``register`` block, otherwise the
+  data will be rejected during account creation.
+
+- Only the First Name, Last Name, Email, and Password fields are shown by
+  default (the middle name and username fields are no longer shown be default).
```
rdegges: *by

docs/registration.rst
```diff
@@ -38,40 +38,95 @@ we will cover them in detail below:
...
-* lastName
+* givenName
+* surname
 * email
 * password
 
-While email and password will always be required (you'll get an API error if
-you omit them), you may not need to require first and last name.  These
-can be configured, and we'll cover that in the next section.
+While email and password will always be required, you may not need to require
```
rdegges: *may not = do not

lib/config.yml
```diff
@@ -27,51 +27,51 @@ web:
...
         placeholder: "Password"
         required: true
         type: "password"
-      passwordConfirm:
-        name: "passwordConfirm"
-        placeholder: "Confirm Password"
-        required: false
-        type: "password"
     fieldOrder:
-      - "username"
```
rdegges: I think username should be listed first. Above givenName / surname. It makes more sense this way in most registration pages.

lib/config.yml
```diff
@@ -27,51 +27,51 @@ web:
     enabled: false
     uri: "/register"
     nextUri: "/"
-    autoAuthorize: false
+    autoLogin: false
```
rdegges: :( Can't we just set to true anyway? I know we can't, but it's sooo tempting.

rdegges:
Made some comments -- but it looks good to me. +1

robertjd:
*boosh* FIX EVERYTHING

thanks @rdegges for all the feedback!

typerandom:
@robertjd Awesome!

Issue #180, opened 2015-11-02
Bug: express-stormpath is incompatible with node streams (request.pipe and http-proxy)


baer:
express-stormpath is crashing my application when I `POST` to a proxied route in express regardless of whether or not it is protected by`stormpath.loginRequired`. Removing the `app.use` in my express config fixes the problem. 

I will keep looking into the issue and will continually update as I find things.

```js
$.ajax({
  url: 'api/aRouteToPostTo',
  type: 'POST',
  data: savableOrder,
  dataType: 'json'
});
```

```js
app.all('/api/*', function(req, res) {
  req.url = req.url.slice(5, req.url.length);
  req.url = '/myApiRootUrl/' + req.url;

  var auth = 'Basic ' + new Buffer(config.proxyUsername + ':' + config.proxyPassword).toString('base64');
  req.headers.authorization = auth;

  proxy.web(req, res, {
    target: 'http://' + config.proxyHost + ':' + config.proxyPort
  });
});
```

```js
app.use(stormpath.init(app, {
  website: true,
  client: {
    apiKey: {
      id: config.stormpath.id,
      secret: config.stormpath.secret
    }
  },
  application: {
    href: config.stormpath.url
  },
  web: {
    register: {
      fieldOrder: ['givenName', 'surname', 'email', 'password', 'passwordConfirm']
    }
  }
}));
```
```
POST /api/aRouteToPostTo - - ms - -
Error: socket hang up
    at createHangUpError (_http_client.js:209:15)
    at Socket.socketCloseListener (_http_client.js:241:23)
    at emitOne (events.js:82:20)
    at Socket.emit (events.js:169:7)
    at TCP._onclose (net.js:469:12)
---------------------------------------------
    at Array.stream (/Users/xxxxxx/dev/stormpath-app/node_modules/http-proxy/lib/http-proxy/passes/web-incoming.js:134:14)
    at ProxyServer.<anonymous> (/Users/xxxxxx/dev/stormpath-app/node_modules/http-proxy/lib/http-proxy/index.js:80:21)
    at /Users/xxxxxx/dev/stormpath-app/src/server/routes.js:52:11
    at Layer.handle [as handle_request] (/Users/xxxxxx/dev/stormpath-app/node_modules/express/lib/router/layer.js:95:5)
    at next (/Users/xxxxxx/dev/stormpath-app/node_modules/express/lib/router/route.js:131:13)
[...]

PR #197, opened 2015-11-11
[POC - DO NOT MERGE] Prevent body-parsing on non-stormpath managed routes

### LINKED ISSUE #194: Body-parser conflicts if configured outside this library
The login handler accepts JSON and form-encoded posts.  We use the body-parser module to parse these.  When a social login is performed the posted object has nested properties:

```
{
  providerdData: {
    providerId: '',
    code: ''
  }
}
```

To read these types of bodies, we use the `extended: true` option of the body-parser library.

However, if the developer has configured the body-parser library outside of our library like this (meaning that this middleware function is place before our router):

```
app.use(bodyParser.urlencoded({ extended: false }));
```

The body will be parsed without the extended fields, and our login handler will fail and return "invalid username or password".  This is a nuance of the body-parser library, as it won't re-parse a body if it's already done so:

https://github.com/expressjs/body-parser/blob/master/lib/types/urlencoded.js#L80

The workaround is to tell the developer to ensure that the option is set to `true`.  But it would be nice to find a body parsing solution that isn't affected by edge cases that arise from configuration that is outside of this library

baer:
After investigating a bit, it looks like there are 3 ways to solve #180 One of which will close #194 as well:

1. Rather than installing bodyParser globally, install a middleware that only applies the body parser to the routes you care about. That is what I did for this PR since it is the most minimally invasive (read: quick and dirty).
2. Just use the bodyParser as route middleware instead of global middleware. e.g. `app.get(/"login", bodyParser, myLoginHandler)`. This one is nice since it has no global middleware installed and is very explicit but the disadvantage is more boilerplate each time you write a route.
3. Remove bodyParser altogether and replace it with [body](https://www.npmjs.com/package/body) or [raw-body](https://www.npmjs.com/package/raw-body). The HUGE advantage here is that you both shrink your footprint on the integrator's app and you remove the dependence on the integrator using the specific body parser config as mentioned in #194. This will close both #180 and #194 but it will require a bit of rework in the app.

Since apps like mine that rely on [http-proxy](https://github.com/nodejitsu/node-http-proxy) or node streams for requests cannot use this lib at all unless some solution is in place I'd prefer to start with 1 or 2 since it's super quick then implement 3 if it is possible.

For this PR, I didn't know how you would want to enumerate routes so I left the array empty. If you can help me figure out how to get the route list in there (and make sure it works with query params) then you can just use this.

### DIFF
--- a/lib/stormpath.js
+++ b/lib/stormpath.js
@@ -72,12 +72,15 @@ module.exports.init = function (app, opts) {
   var client = initClient(app, opts);
 
   // Parse the request body.
-  app.use(function(req,res,next) {
-    // TODO: I do not know where/what all of these are
-    var stormpathRoutes = [];
+  router.use(function(req, res, next) {
+    var exemptRoutes = ["/api/"];
 
-    if(stormpathRoutes.indexOf(req.path) > -1) {
-      router.use(bodyParser.urlencoded({ extended: true }));
+    var skipBodyParsing = exemptRoutes
+      .map(function (exemption) { return req.path.match(exemption) !== null; })
+      .reduce(function (value, memo) { return value || memo; }, false);
+
+    if (skipBodyParsing) {
+      bodyParser.urlencoded({ extended: true })(req, res, next);
     }
     else {
       return next();
--- a/lib/stormpath.js
+++ b/lib/stormpath.js
@@ -72,7 +72,17 @@ module.exports.init = function (app, opts) {
   var client = initClient(app, opts);
 
   // Parse the request body.
-  router.use(bodyParser.urlencoded({ extended: true }));
+  app.use(function(req,res,next) {
+    // TODO: I do not know where/what all of these are
+    var stormpathRoutes = [];
+
+    if(stormpathRoutes.indexOf(req.path) > -1) {
+      router.use(bodyParser.urlencoded({ extended: true }));
+    }
+    else {
+      return next();
+    }
+  });
 
   // Indicates whether or not the client is ready.
   var isClientReady = false;
--- a/lib/stormpath.js
+++ b/lib/stormpath.js
@@ -80,10 +80,9 @@ module.exports.init = function (app, opts) {
       .reduce(function (value, memo) { return value || memo; }, false);
 
     if (skipBodyParsing) {
-      bodyParser.urlencoded({ extended: true })(req, res, next);
-    }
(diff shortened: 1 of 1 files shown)


lib/stormpath.js
```diff
@@ -72,7 +72,17 @@ module.exports.init = function (app, opts) {
   var client = initClient(app, opts);
 
   // Parse the request body.
-  router.use(bodyParser.urlencoded({ extended: true }));
+  app.use(function(req,res,next) {
+    // TODO: I do not know where/what all of these are
```
baer: Halp! :)

robertjd:
Thanks @baer !  This is really useful.  We're pretty busy with the existing backlog, but we'll look into this as soon as possible.

baer:
@robertjd I really need some form of this PR to go in since stormpath is unusable to me (any anybody else who proxys/manipulates POSTs/PUTs in their app) until it goes in. I understand that you have an existing backlog but I'm happy to do the work required provided we have a solution that we can agree on.

I have an even less invasive idea if you're open to the PR: What if you allow a "disableBodyParsing" option that allows a user to pass in routes that they don't want to be parsed. That way you don't even have to enumerate all of your routes.

baer:
I updated the PR to do what I suggested but have not wired it into the options object yet. If you would be open to merging this then I will wire up the rest.

robertjd:
Hey @baer - I found some time to look into this today :)

I prefer to go with (3), and just replace body parser entirely.  I've done that in this branch:

https://github.com/stormpath/express-stormpath/tree/replace-body-parser

Can you give that branch a try?  I'd like to know if it solves your proxy problem.  Thanks!

baer:
@robertjd - I'd be happy to. I'll point my project at your branch and let you know before Monday. You da best! :zap:!

baer:
@robertjd - It worked like a charm but I was wondering why you didn't just have your `jsonParser` and `formEncodedParser` return the two `body-parser` methods you replace. Both of the `body-parser` related bugs mentioned above are a result of using `body-parser` as app-wide middleware and not a problem with the lib itself. I see that you have the `body-parser` methods you are replacing in comments above each helper; would this branch work just as well if you returned the midleware that the `body-parser` factory creates?

typerandom:
Closing this in favour for #256.

PR #198, opened 2015-11-12
Supporting local YAML overrides of framework options

robertjd:
@typerandom I noticed today that it was not possible to override the default framework configuration by placing a `config.yaml` file in the root of a project which depends on `express-stormpath`

This fixes the problem, but do you think this is the right way to do it?  The underlying `stormpath.configParser` *is* looking for the file in question, but I think it framework options get overriden when `epxress-stormpath` applies it's default configuration file.  Thus needing to re-apply the custom file, as you see in this commit.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -23,5 +23,8 @@ module.exports = function (config) {
   configLoader.add(new configStrategy.EnrichIntegrationFromRemoteConfigStrategy(ClientFactory));
   configLoader.add(new configStrategy.EnrichIntegrationConfigStrategy());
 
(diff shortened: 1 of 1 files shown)


typerandom:
I don't know that `config.yml` should be supported. I guess what you're looking for is `stormpath.yml`. Have you tried that?

I.e. see: https://github.com/stormpath/stormpath-sdk-node/blob/master/lib/configLoader.js#L35

robertjd:
Oops :) yeah it works with stormpath.yml

PR #207, opened 2015-11-20
Fixing the spaRoot option

robertjd:
We were looking at the wrong option, and weren’t binding in all cases where we need to.


rdegges:
+1

PR #215, opened 2015-11-30
Fix rendering of errors during Facebook login attempt

robertjd:
We want to send the error message directly to the browser, we don’t have a specific error view for this case (that was from the 1.x version)


rdegges:
+1

PR #223, opened 2015-12-07
Updates to social login documentation

robertjd:
* The default callback URI for google is now shown correctly

* Adding notes about how to use the JSON API for social login w/ pop-up based social login flows


rdegges:
+1

PR #229, opened 2015-12-11
Fix: secure cookie properties

robertjd:
The reference to `https` was undefined, which meant that we would pass undefined for the secure value (which meant that the secure value was never set to true).  This PR fixes that.


rdegges:
+1

PR #235, opened 2015-12-15
Use field label when rendering field validation errors

robertjd:
This is more user-friendly than "givenName", etc.

To verify:

Use our [sample project](https://github.com/stormpath/express-stormpath-sample-project) and visit the registration form.  Use your web inspector to remove the `required` attribute from the "First Name" field, so that HTML form validation does not prevent you from submitting the form.  Submit the form and you should see the error "First Name required"

This should go in a minor release, and the upgrade log should note that the error messages are changing (as some may be depending on the errors, particularly if using the JSON responses)


typerandom:
Works and looks good! :+1:

PR #244, opened 2015-12-18
Successful user registration via IdSite results in error msg to user …

j0nesin:
…: Unknown ID site result status: REGISTERED


robertjd:
Thanks for the PR!  There is one modification we need to make: the session should only be created if `stormpath.web.register.autoLogin` is set to true.

j0nesin:
Added check for autoLogin.

rdegges:
We'll be getting to this shortly. Thanks for waiting. Just prioritized this with the team.

PR #252, opened 2015-12-26
Fix: Ability to disable web features

### LINKED ISSUE #250: Can't disable web features
If I use the `website: true` option, but also try to disable a specific feature, the feature is not disabled.

To reproduce, run this code, and see that the final config object has the value of `true` (not `false`) for `web.register.enabled`:

```
var express = require('express');
var stormpath = require('express-stormpath');

var app = express();

app.use(stormpath.init(app, {
  website: true,
  web: {
    register: {
      enabled: false
    }
  }
}));

app.on('stormpath.ready',function () {
  console.log(app.get('stormpathConfig').web.register);
});

```

@typerandom please prioritize this one.

typerandom:
Enables the ability to set web features while still having the website option turned on.

Fixes #250.


rdegges:
+1, but stop working on the holiday! <3

PR #254, opened 2015-12-27
Fix: Load express version from application modules

### LINKED ISSUE #239: Express version reporting is incorrect
We're adding an express version to our user agent string, here:

https://github.com/stormpath/express-stormpath/blob/master/lib/stormpath.js#L26

However, that is going to require the version of express that is depended on by this library in package.json (not the version of express that the developer is using). The UA string is not telling us what version of express was used to create the app that was passed into our init() function.

As such, unless we can find a way to know what version of express has been used to create the app (I don't know how?), I think we should remove this from the UA because it's useless.  We should also move express to `devDependencies`, as there is no reason to have this as a prod dependency (we only use express directly in our tests).

Ping @typerandom @rdegges

typerandom:
Currently the Express version that we report in our user agent is the one that we have loaded in our package (`express-stormpath`). I.e. it might not be the same that the user is actually using.

Therefore I suggest this fix, so that we resolve the package ourselves by first looking in the path of the running process and its node_modules directory, then if that doesn't exist we fall back to the global node_modules directory.

If we are able to find a module in either one of those directories, then we simply load the package.json file from there and get the version.

Fixes #239

### DIFF
--- a/lib/helpers/get-app-module-version.js
+++ b/lib/helpers/get-app-module-version.js
@@ -1,0 +1,60 @@
+'use strict';
+
+var fs = require('fs');
+var path = require('path');
+
+var cachedModulePaths = {};
+var globalModulesPath = process.env.NODE_PATH;
+var appFilePath = require.main.filename;
+var appModulesPath = path.join(path.dirname(fs.realpathSync(appFilePath)), 'node_modules');
+
+function getApplicationModulePath(name) {
+  if (name in cachedModulePaths) {
+    return cachedModulePaths[name];
+  }
+
+  var result = false;
+
+  var modulePaths = [
+    path.join(appModulesPath, name)
+  ];
+
+  if (globalModulesPath) {
+    modulePaths.push(path.join(globalModulesPath, name));
+  }
+
+  while (modulePaths.length) {
+    var modulePath = modulePaths.shift();
+    if (fs.existsSync(modulePath)) {
+      result = modulePath;
+      break;
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


lib/helpers/get-app-module-version.js
```diff
@@ -0,0 +1,59 @@
+'use strict';
+
+var fs = require('fs');
+var path = require('path');
+
+var cachedModulePaths = {};
+var appModulesPath = path.join(path.dirname(fs.realpathSync(__filename)), 'node_modules');
```
timothyej: This still just gets the modules for this lib, and not the application. In my case, it's currently looking in: 

`/Users/timothy/Projects/stormpath/my-react-app/node_modules/express-stormpath/lib/helpers/node_modules/express`

But it should be looking in:

`/Users/timothy/Projects/stormpath/my-react-app/node_modules/express`

By changing this line to the following, I made it work:

```javascript
var appModulesPath = path.join(path.dirname(fs.realpathSync(require.main.filename)), 'node_modules');
```

But I don't know how safe it is to use `require.main.filename`. You can read more about it here: [Determine project root from a running node.js application](http://stackoverflow.com/questions/10265798/determine-project-root-from-a-running-node-js-application)

typerandom:
@timothyej My bad! This is absolutely correct, `require.main.filename` is the path that we're supposed to be using. :+1: I'll fix that.

typerandom:
@timothyej This has been changed, please try it again.

timothyej:
:+1:

PR #255, opened 2015-12-28
Refactor: Rename serveAngular as addSpaRoute

typerandom:
Small internal refactor. Simply make the use of `serveAngular()` clearer by renaming it to `addSpaRoute()`.

### DIFF
--- a/lib/stormpath.js
+++ b/lib/stormpath.js
@@ -113,7 +113,7 @@ module.exports.init = function (app, opts) {
       next();
     }
 
-    function serveAngular(path) {
+    function addSpaRoute(path) {
       router.get(path, function (req, res) {
         res.sendFile(web.spaRoot);
       });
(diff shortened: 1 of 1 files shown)


robertjd:
This new name feels strange to me, perhaps something like "attachSpaHandler"?  FYI, this mechanism will get re-factored again when we re-visit our SPA strategy, per the new spec that Les and I worked out a few weeks ago:

https://github.com/stormpath/stormpath-framework-spec/blob/master/single-page-apps.md

typerandom:
@robertjd Personally I feel that `addSpaRoute` is more explicit. Since "handler" for me could be anything. And I prefer `add` since it's shorter. Any specific reason it feels strange for you? The function would look like `addSpaRoute(handler)`, so for me that feels pretty descriptive.

And +1 for "SPA" spec, I look forward to when we can start implementing that :)

timothyej:
How about `attachSpaRoute()` or `setupSpaRoute()`?

typerandom:
@timothyej Does that add different value than `addSpaRoute`?

timothyej:
Since we can't agree on anything at this moment, and `addSpaRoute` is still better than `serveAngular`, I'm going to merge this :+1:

PR #256, opened 2015-12-28
Fix: Remove body-parser dependency

### LINKED ISSUE #194: Body-parser conflicts if configured outside this library

typerandom:
Removes the `body-parser` dependency and instead uses the `body` and `qs` module to parse JSON and query string encoded request bodies.

Fixes #194 and #180

### DIFF
--- a/lib/helpers/body-parser.js
+++ b/lib/helpers/body-parser.js
@@ -1,0 +1,55 @@
+'use strict';
+
+var qs = require('qs');
+var bytes = require('bytes');
+var anyBody = require('body/any');
+var formBody = require('body/form');
+
+var defaultSizeLimit = '200kb';
+
+function queryStringParser(text, callback) {
+  callback(null, qs.parse(text));
+}
+
+function handleBodyFn(options, bodyFn) {
+  options = options || {};
+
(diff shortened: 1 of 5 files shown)


timothyej:
Nice! :+1:

PR #271, opened 2016-01-19
Fix: Handle id site registered status

typerandom:
Handles the ID site `registered` result status.

Fixes #244.

### DIFF
--- a/lib/controllers/id-site-verify.js
+++ b/lib/controllers/id-site-verify.js
@@ -24,16 +24,26 @@ module.exports = function (req, res) {
       return res.status(500).end(err.toString());
     }
 
-    if (result.status === 'LOGOUT') {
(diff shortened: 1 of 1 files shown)


rdegges:
+1 the code looks perfect -- wanna fix the broken build though? Looks like there's some whitespace issue.

robertjd:
@rdegges can you do a full UAT with our sample application to ensure that the problem is resolved?  Thanks!

typerandom:
@rdegges Thanks man! It's fixed now. For some reason when using Sublime I sometimes insert characters that look like spaces but have some totally other character code. It has happened to me on several occasions. Quite annoying. So I need to get better at running lint right before pushing:)

rdegges:
Working.

PR #276, opened 2016-01-21
Fix: Body parser middleware issues

### LINKED ISSUE #264: New body-parser helper resets the body, breaks middleware that expect streams
Related to #256.

I'm using [rocky](https://github.com/h2non/rocky/) to proxy some of the routes of my express.js application, and using the `express-stormpath` library breaks the request body because the body-parser helper is "forcing" a default body of an empty object. The proxy expects the default body of an express request(a string or a buffer), so resetting the body to an empty object is breaking the request flow.

To fix this issue I commented out the following in `stormpath.js` in the sdk itself: 
```js 
router.use(bodyParser.forceDefaultBody())
```

Not sure if this breaks some other functionality(doesn't seem to) but I would say modifying the req.body on every request is not such a great idea anyway, so maybe there is a better way to handle this within the library?

### LINKED ISSUE #289: POST request hangs if body-parser is configured
If the `body-parser` module is configured with express as `app.use(bodyParser.urlencoded({ extended: true }));` then any POST requests handled by `express-stormpath` hangs.

#### How to reproduce

1. Create a new server using [this gist](https://gist.github.com/typerandom/b1178002a81751b82923).
2. Start the server.
3. Navigate to [/register](http://localhost:3000/register).
4. Sign up for a new account.

#### Actual

Page hangs indefinitely.

#### Expected

Page is redirected to [/login?status=verified](http://localhost:3000/login?status=verified). If you have email verification turned on, then it should redirect to [/login?status=unverified](http://localhost:3000/login?status=unverified).

typerandom:
Fixes the default body parser object issue (#264) by only forcing a default body on requests that we serve and fixes the body parser hang issue (#276) by only parsing the body when it hasn't already been parsed.

##### Default body parser (fix #264)

1. Checkout this branch (`fix-body-parser-middleware`).
2. Execute `$ npm link` so that we can link to it.
3. Create a new directory and configure a new Node.js server according to [this gist](https://gist.github.com/typerandom/11d2d428e5565ef7a57d).
4. Execute `$ npm link stormpath-express` so that we're linking to our `fix-body-parser-middleware` branch.
5. Start the server.
6. Navigate to [http://localhost:3000/register](http://localhost:3000/register).
7. Create a new account.
8. Navigate to [http://localhost:3000/forgot](http://localhost:3000/forgot).
9. Request a new password.
10. Navigate to [http://localhost:3000/login](http://localhost:3000/login).
11. Login with the credentials you provided in step 7.
12. Navigate to [http://localhost:3000/logout](http://localhost:3000/logout).
13. Check the console for errors.
14. If no errors in console and steps 6 to 9 went smooth, then \o/.

##### Body parser hanging (fix #289)

1. Create a new directory and configure a new Node.js server according to [this gist](https://gist.github.com/typerandom/b1178002a81751b82923).
2. Execute `$ npm link stormpath-express` so that we're linking to our `fix-body-parser-middleware` branch.
3. Start the server.
4. Navigate to [http://localhost:3000/register](http://localhost:3000/register).
5. Create a new account.
6. Page is redirected to [/login?status=verified](http://localhost:3000/login?status=verified). If you have email verification turned on, then it should redirect to [/login?status=unverified](http://localhost:3000/login?status=unverified).
7. Open up your server and remove the `app.use(bodyParser.urlencoded({ extended: true }));` line.
8. Repeat steps 3-6.

Fixes #264, #289

### DIFF
--- a/lib/helpers/body-parser.js
+++ b/lib/helpers/body-parser.js
@@ -18,6 +18,11 @@ function handleBodyFn(options, bodyFn) {
   options.limit = bytes.parse(options.limit || defaultSizeLimit);
 
   return function (req, res, next) {
+    // If the body is already parsed, by e.g. body-parser, then skip parsing.
+    if (req.body !== undefined) {
+      return next();
+    }
+
     bodyFn(req, res, options, function (err, parsedBody) {
       req.body = parsedBody || req.body || {};
       next();
--- a/lib/stormpath.js
+++ b/lib/stormpath.js
@@ -116,40 +116,47 @@ module.exports.init = function (app, opts) {
       });
     }
 
+    function addGetRoute(path, controller) {
+      router.get(path, bodyParser.forceDefaultBody(), controller);
+    }
+
+    function addPostRoute(path, controller, options) {
+      router.post(path, bodyParser.formOrJson(options), controller);
+    }
+
     router.use(localsMiddleware);
     router.use(cookieParser());
-    router.use(bodyParser.forceDefaultBody());
 
-    router.get('/spa-config', controllers.socialProviders);
+    addGetRoute('/spa-config', controllers.socialProviders);
 
     if (web.idSite.enabled) {
-      router.get(web.idSite.uri, controllers.idSiteVerify);
+      addGetRoute(web.idSite.uri, controllers.idSiteVerify);
     }
 
     if (web.register.enabled) {
       if (web.idSite.enabled) {
-        router.get(web.register.uri, controllers.idSiteRedirect({ path: web.idSite.registerUri }));
+        addGetRoute(web.register.uri, controllers.idSiteRedirect({ path: web.idSite.registerUri }));
       } else {
         if (web.spaRoot) {
           addSpaRoute(web.register.uri);
         } else {
-          router.get(web.register.uri, controllers.register);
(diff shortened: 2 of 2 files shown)


robertjd:
HI @typerandom , thanks for the fix!  Unfortunately this PR still has an issue: the POST handler will hang if the developer has configured the `bodyParser.urlencoded({ extended: true })` middleware before ours.

You can see this problem by looking at this sample application, on this branch:

https://github.com/stormpath/express-stormpath-angular-sample-project/tree/broken-registration-post

http://localhost:3000/register

When you try to POST to the /register route from the front-end, it hangs.

timothyej:
@typerandom @robertjd I forgot to mention it yesterday while I was reviewing this. But this isn't a problem with this PR, but a general problem. Same issue can be reproduced on `master` :(

robertjd:
I've verified that this fixes #289 with your second test-case in the issue description.  I have not explicilty tried to use the rocky module to verify that #264 is fixed, but I have verified that our module is no longer setting a default req.body object.  As such I'm going to merge this as it appears to fix the problems that are causing trouble.  Thanks @typerandom  for taking this one!

rdegges:
Great job guys!

PR #291, opened 2016-01-28
Upgrade lodash

robertjd:
Easy API name changes


timothyej:
Seems like these two methods are the only ones that are being used from `lodash`. Maybe we should consider getting rid of this dependency?

I will merge this either way :+1:

robertjd:
IMO we should be using it more, and using it to replace code bloat like these examples:

https://github.com/stormpath/express-stormpath/blob/master/lib/helpers/get-form-model.js#L22-L35

https://github.com/stormpath/express-stormpath/blob/master/lib/controllers/social-providers.js#L15-L25

Issue #230, opened 2015-12-11
Server Crash when Disabling Refresh Tokens


rdegges:
If I create a Stormpath Application, and disable refresh tokens in the dashboard, eg:

<img width="1057" alt="screenshot 2015-12-11 10 18 32" src="https://cloud.githubusercontent.com/assets/90247/11751804/ac7aa1e2-9ff0-11e5-8786-4a0759281cd8.png">

Then when I go to register a user using the default registration page stuff, my server crashes:

```
nodemon server.js
[nodemon] 1.8.1
[nodemon] to restart at any time, enter `rs`
[nodemon] watching: *.*
[nodemon] starting `node server.js`
/Users/rdegges/Dropbox/Code/testing/timeout/node_modules/express-stormpath/lib/helpers/create-session.js:31
    expires: new Date(passwordGrantAuthenticationResult.refreshToken.body.exp * 1000),
                                                                    ^

TypeError: Cannot read property 'body' of undefined
    at Object.module.exports [as createSession] (/Users/rdegges/Dropbox/Code/testing/timeout/node_modules/express-stormpath/lib/helpers/create-session.js:31:69)
    at /Users/rdegges/Dropbox/Code/testing/timeout/node_modules/express-stormpath/lib/controllers/register.js:247:21
    at /Users/rdegges/Dropbox/Code/testing/timeout/node_modules/express-stormpath/node_modules/stormpath/lib/oauth/password-grant.js:38:7
    at onCreateResourceRequestResult (/Users/rdegges/Dropbox/Code/testing/timeout/node_modules/express-stormpath/node_modules/stormpath/lib/ds/DataStore.js:359:14)
    at Request.onRequestResult [as _callback] (/Users/rdegges/Dropbox/Code/testing/timeout/node_modules/express-stormpath/node_modules/stormpath/lib/ds/RequestExecutor.js:110:7)
    at Request.self.callback (/Users/rdegges/Dropbox/Code/testing/timeout/node_modules/express-stormpath/node_modules/request/request.js:198:22)
    at emitTwo (events.js:87:13)
    at Request.emit (events.js:172:7)
    at Request.<anonymous> (/Users/rdegges/Dropbox/Code/testing/timeout/node_modules/express-stormpath/node_modules/request/request.js:1035:10)
    at emitOne (events.js:82:20)
[nodemon] app crashed - waiting for file changes before starting...
```

We should fix this so that we aren't assuming the refresh token value is defined.

[...]

PR #296, opened 2016-02-01
Feature: Replace ID site token with Stormpath token

### LINKED ISSUE #251: Replace "idSiteSession" cookie with Oauth tokens
We have new features in the Stormpath API that allow us to:

* Exchange an ID Site result (the jwt that is delivered to the callback url) for an access token + refresh token pair.  Simply use `grant_type=stormpath_token` and `token=<jwt>` on the application's oauth token endpoint, where `<jwt>` is the jwt result from ID Site.

* For social login, use `grant_type=stormpath_token` and  `token=<jwt>`, where `<jwt>` has the same format as the id site result jwt, but the `iss` is the stormpath application that has serviced the social login attempt.  This jwt needs to be signed with a tenant api key.

typerandom:
Replaces the ID site token with a Stormpath OAuth access/refresh token.

Requires https://github.com/stormpath/stormpath-sdk-node/pull/362 to be merged.

Fixes #251, #230.


robertjd:
Hey @typerandom , looking at the diff this is getting closer.  However I'm running into the same problem that is seen in the travis tests: the ID Site callback blows up in `create-stormpath-session.js` because `helpers` is an empty object, which is pretty strange.  Are you not seeing this on your machine?  This will require some investigation.

Other notes:

* In the Node SDK: `OAuthStormpathTokenGrantAuthenticator` should not have Grant in the name.  This will keep it consistent with the SDK spec.

* The way this is written, you always create a new JWT before making the `grant_type=stormpath_token` request.  This *is* necessary for the social login flow, but not the ID Site callback - that JWT can be passed directly as `stormpath_token`.  As such, a ID Site callback should look like this:
 * Use the `StormpathAssertionAuthenticator` to validate the JWT from the callback.  If there is an error, show it to the user
 * If no errors, pass the JWT to the `OAuthStormpathTokenAuthenticator`
 * Take the result and create cookies

Through this refactor, you should be able to remove the call to `application.handleIdSiteCallback`, which is what we want.

typerandom:
@robertjd 

Thanks!

I have fixed this PR according to your feedback. By old habits I've done a rebase/squash. Remember that you didn't like that when reviewing... Will try to think of not doing that in the future:)

Also updated https://github.com/stormpath/stormpath-sdk-node/pull/362.

robertjd:
Looking better, however we lost most of our feature support for ID Site:

* Logout needs to be redirected through ID Site, see [Logging Out of ID Site with REST](https://docs.stormpath.com/rest/product-guide/latest/idsite.html#step-3-optional-logging-out-of-id-site-with-rest).  
* Registration callbacks from ID Site need to continue to respond to the `REGISTERED` status in the token, and call the post-registration callbacks if they exist.
* Login callbacks from ID Site need to continue to respond to the `AUTHENTICATED` status, and call post-login callbacks if they exist.

Back to @typerandom

typerandom:
@robertjd 

Okay. I assumed when you said that we would remove `handleIdSiteCallback()` the thought was for us to handle it in some other way. But I'll fix that.

Just some things that come up...

Did we not handle the post registration/login callbacks and the ID site logout before? I'm adding it now, but cannot see that we did...

typerandom:
@robertjd I have updated the PR according to your comments. What do you think?

lib/controllers/id-site-verify.js
```diff
@@ -18,31 +21,78 @@ module.exports = function (req, res) {
...
+        }
+
+        helpers.createSession(authenticationResult, authenticationResult.account, req, res);
+
+        callback();
+      });
+    }
+
+    function handleAuthRequest(type, callback) {
+      var handler = config['post' + type + 'Handler'];
```
robertjd: I generally avoid dynamic function names, as it makes the code harder to read (IMO), and I'm pretty sure it makes the job difficult for static analyzers.

lib/controllers/id-site-verify.js
```diff
@@ -18,31 +21,78 @@ module.exports = function (req, res) {
...
+
+      if (handler) {
+        handler(req.user, req, res, function () {
```
robertjd: req.user will never be available because it's not set until `helpers.createSession` is called, inside of `authenticateToken`

lib/controllers/id-site-verify.js
```diff
@@ -18,31 +21,78 @@ module.exports = function (req, res) {
...
+    function authenticateToken(callback) {
+      var stormpathTokenAuthenticator = new stormpath.OAuthStormpathTokenAuthenticator(application);
+
+      stormpathTokenAuthenticator.authenticate({ stormpath_token: stormpathToken }, function (err, authenticationResult) {
+        if (err) {
+          logger.info('During an IdSite login attempt, we were unable to create a Stormpath session.');
+          return res.status(err.status || 400).json(err);
+        }
+
+        helpers.createSession(authenticationResult, authenticationResult.account, req, res);
```
robertjd: This needs to pass the expanded account, not the account object that hangs off the authentication result - that is merely a reference that needs to be resolved with a getter.

robertjd:
Hey @typerandom , please see new issues via inline comments.  The login and registration handlers are called now, but with the wrong parameters.  Please confirm that you are trying these features with a demo project, as you develop them?  Off all the issues that I've had to pass back, I've found them while configuring our [sample project](https://github.com/stormpath/express-stormpath-sample-project) to use the ID Site and post-handler features that are documented in the library documentation.

lib/controllers/id-site-verify.js
typerandom: Yeah, I fully agree with that reasoning. Will change that.

lib/controllers/id-site-verify.js
typerandom: Yeah sorry, I forgot to try out the post login/registration features. Copied the use from another function, and forgot about the state in which `req.user` was in. Should have been obvious, but something I clarely overlooked. So thanks for noticing. Will fix that!

typerandom:
@robertjd Thanks! I have been trying the ID site features with my own demo project, but I never tried the post login/registration because I forgot about those (doh!). And the mistake with `req.user` there was from copying the usage from another method and forgetting the state in which `req.user` was in. But as you point out, it's kind of obvious why that isn't working.

lib/helpers/create-stormpath-session.js
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Creates a Stormpath JWT, exchanges it for an OAuth token, and stores it in a cookie.
+ *
+ * @param  {Object} account              Account to create a session for.
+ * @param  {Object} req                  Express HTTP request.
+ * @param  {Object} res                  Express HTTP response.
+ * @param  {function} callback           Callback to call once the token cookie(s) has been set.
+ */
+module.exports = function createStormpathToken(account, req, res, callback) {
```
timothyej: The name of this file is `create-stormpath-session.js` and is exported as `createStormpathSession` in `index.js`. Is there any reason why the function is called `createStormpathToken` here?

lib/controllers/facebook-login.js
```diff
@@ -60,21 +60,27 @@ module.exports = function (req, res) {
 
       res.locals.user = expandedAccount;
       req.user = expandedAccount;
-      helpers.createIdSiteSession(req.user, req, res);
 
-      var nextUrl = url.parse(req.query.next || '').path || (resp.created ? config.web.register.nextUri : config.web.login.nextUri);
+      helpers.createStormpathSession(req.user, req, res, function (err) {
```
timothyej: The callbacks for all the social providers looks very similar. Actually the whole controllers looks very similar to each other. Any reason why they are not just one generic controller?

*I know this is not your fault, I'm just curious :)*

lib/helpers/create-stormpath-session.js
typerandom: Just a typo from a refactor. Thanks for noticing:)

test/controllers/test-id-site.js
```diff
@@ -214,7 +214,7 @@ describe('id site', function () {
   it('should create a session on id site callback', function (done) {
     getIdSiteCallbackUrl(host, config, app, accountData, function (url) {
       request(url).get('')
-        .expect('Set-Cookie', /idSiteSession=https:\/\/api.stormpath.com\/v1\/accounts\/.*/)
+        .expect('Set-Cookie', /access\_token=.*refresh\_token=.*/)
```
timothyej: You don't have to escape `_` in regex.

lib/controllers/facebook-login.js
typerandom: I'd say it's out of scope for this PR. But Yeah, the Facebook and Google providers could probably be refactored to one. LinkedIn does some things slightly different, but that could probably also be refactored:)

test/controllers/test-id-site.js
typerandom: Yeah, I thought that represented a character class, but I must have confused that with something else. Fixing :)

typerandom:
@robertjd I have updated the PR with regards to your comments. The `postLoginHandler` is working, but I'm not getting the `postRegistrationHandler` to work. Seems to be something wrong in the callback. Do you see anything immediately wrong in my fix?

typerandom:
@robertjd Apparently `autoLogin` is not being set. I thought that was only if I had email verification activated... Which I have disabled. But I'll investigate what is not being set.

typerandom:
@robertjd Okay, it was just my assumption that we automatically set `autoLogin` (which we don't). If I set that manually then I get the callback. But, when I try to handle it I get a new error...

http://localhost:3000/idSiteResult?jwtResponse=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJodHRwczovL21hc3NpdmUtYm9iY2F0LmlkLnN0b3JtcGF0aC5pbyIsInN1YiI6Imh0dHBzOi8vYXBpLnN0b3JtcGF0aC5jb20vdjEvYWNjb3VudHMvNTg3V1Z4Rm5leVExdmRtSFdCY0RZWiIsImF1ZCI6IjRWUE5XRE5GMDlLMTNUUDNEQTlDQldQWDQiLCJleHAiOjE0NTQ2MDAwNzcsImlhdCI6MTQ1NDYwMDAxNywianRpIjoiNThBQW55T2w1YUp6MVZuUGFxOXVzZCIsImlydCI6IjUwZGM3M2YwLTVlN2QtNGFiZi04ZDVkLTExODhjZjUzYjkzNCIsInN0YXRlIjoiIiwiaXNOZXdTdWIiOnRydWUsInN0YXR1cyI6IlJFR0lTVEVSRUQiLCJjYl91cmkiOiJodHRwOi8vbG9jYWxob3N0OjMwMDAvaWRTaXRlUmVzdWx0In0.xsjDsUNC5sE-9FGIPja6hS2a-F4chd4lMcBu2qDc1Js

##### jwtResponse JWT:

```json
{
  "iss": "https://massive-bobcat.id.stormpath.io",
  "sub": "https://api.stormpath.com/v1/accounts/587WVxFneyQ1vdmHWBcDYZ",
  "aud": "4VPNWDNF09K13TP3DA9CBWPX4",
  "exp": 1454600077,
  "iat": 1454600017,
  "jti": "58AAnyOl5aJz1VnPaq9usd",
  "irt": "50dc73f0-5e7d-4abf-8d5d-1188cf53b934",
  "state": "",
  "isNewSub": true,
  "status": "REGISTERED",
  "cb_uri": "http://localhost:3000/idSiteResult"
}
````

##### Page error JSON:

```json
{
"name": "ResourceError",
"status": 401,
"code": 12002,
"userMessage": "Token is invalid",
"developerMessage": "The ID Site token that was submitted was not authenticated by ID Site. Stormpath can only generate an access token for ID Site tokens that include the status claim of 'AUTHENTICATED', meaning that ID Site authenticated the account with a valid login attempt.",
"moreInfo": "http://docs.stormpath.com/errors/12002",
"message": "HTTP 401, Stormpath 12002 (http://docs.stormpath.com/errors/12002): The ID Site token that was submitted was not authenticated by ID Site. Stormpath can only generate an access token for ID Site tokens that include the status claim of 'AUTHENTICATED', meaning that ID Site authenticated the account with a valid login attempt.",
"error": "invalid_grant"
}
```

Do you have any idea how we should resolve this? Apparently we cannot use the token we get from a registration to authenticate with...

typerandom:
@robertjd I have fixed the logout issue. Also see https://github.com/stormpath/express-stormpath/commit/9820b6ae790c3ac41aca3c7d86ca579d46b60350. I removed the old ID Site cookie auth. What do you think about this? It will of course affect people already logged in. But might be nice to have this cleaned up.

lib/controllers/facebook-login.js
```diff
@@ -60,21 +60,27 @@ module.exports = function (req, res) {
...
+          return res.status(err.status || 400).json(err);
+        }
 
-      if (resp.created && registrationHandler) {
-        registrationHandler(req.user, req, res, function () {
-          res.redirect(302, nextUrl);
-        });
-      } else if (loginHandler) {
-        loginHandler(req.user, req, res, function () {
+        var nextUrl = url.parse(req.query.next || '').path || (resp.created ? config.web.register.nextUri : config.web.login.nextUri);
```
timothyej: A suggestion:

This line is very long and hard to follow. Considering breaking it up to something more readable, like this:

```javascript
var nextUrlConfig = resp.created ? config.web.register.nextUri : config.web.login.nextUri;
var nextUrlQuery = url.parse(req.query.next || '').path;
var nextUrl = nextUrlQuery || nextUrlConfig;
```

typerandom: Yeah... This is not something that I added. I just moved that line because of the code that I added.

timothyej: Oh I see!

robertjd:
At last, I will merge this :)  Thanks @typerandom for all the fixes.  As part of my final review, I added two changes:

* Removing the check for cookies on logout, before redirecting to ID Site.  Even if the cookies have expired locally, we need to redirect to ID Site because another cookie is maintained on api.stormpath.comm for SSO purposes.  That SSO cookie is cleared during the redirect flow

* Implemented our standard json error responder for the ID Site verification flow, so that we don't blast the end-user with a payload of token cruft.  Instead we want to show the concise error message.

PR #304, opened 2016-02-08
Make sure that OAuth callback errors gets handled

### LINKED ISSUE #288: Facebook login breaks if someone declines the `email` permission
We need to fail gracefully on this since it's not guaranteed that we'd get an email. Facebook accounts can be tel: only, for instance. 

See [Facebook's permissions documentation](https://developers.facebook.com/docs/facebook-login/permissions) for more info. 

I initiate the FB login request, and decline the `email` permission: 

screen shot 2016-01-27 at 11 12 20 am

`express-stormpath` responds with barfing out json. :(
 
screen shot 2016-01-27 at 11 12 32 am

Facebook's docs say that we should never rely on the email being there:

screen shot 2016-01-27 at 11 37 49 am

Related: also seems to trigger a similar error on IDSite, although not JSON barf

timothyej:
Fixes #288.

Errors during social login is today displayed as json (see #288). This PR fixes so that the error is instead displayed as an error on the login form. It also shows a more specific error message when email was not provided.

**Please see *Discussion* before reviewing.**

### Discussion

After I've made this, I found out that there are views for displaying social login errors (e.g. `facebook_login_failed.jade`), but they doesn't seem to be used.

However, I found it better to display the error on the login form as it allows the user to try again.

Question is, what solution should I go with? Implementing the error views or use the login form as this PR does? If the latter, I should probably show the registration form instead if the error occurred during signup, which is not possible to know at the moment.

### How to verify

1. **Checkout `master`.**
2. Clone the [express-stormpath-sample-project](https://github.com/stormpath/express-stormpath-sample-project).
3. Link to express-stormpath (`npm link` and `npm link express-stormpath`)
4. Configure it with a Facebook Directory.
5. Try to login with Facebook, but in the login popup, choose to not share your email address.
6. Verify that you get an error response in form of json.
7. **Checkout this branch.**
8. Try to login again.
9. Verify that you get the error on the login form instead (*see screenshot*)
10. Also verify that the error message has been rewritten to state that must you share your email.
11. Browse to <http://localhost:3000/callbacks/facebook?&access_token=sdssdsd> with an invalid access token.
12. Verify that you get another error message.
13. Try the same for Google and LinkedIn.

### Screenshots

screen shot 2016-02-08 at 13 36 44

### DIFF
--- a/lib/controllers/facebook-login.js
+++ b/lib/controllers/facebook-login.js
@@ -38,7 +38,7 @@ module.exports = function (req, res) {
 
   if (!req.query.access_token) {
     logger.info('A user attempted to log in via Facebook OAuth without specifying an OAuth token.');
-    return oauth.errorResponder(req, res, { message: 'access_token parameter required.' });
+    return oauth.errorResponder(req, res, new Error('access_token parameter required.'));
   }
 
   var userData = {
--- a/lib/controllers/facebook-login.js
+++ b/lib/controllers/facebook-login.js
@@ -1,7 +1,9 @@
 'use strict';
 
-var helpers = require('../helpers');
 var url = require('url');
+
+var helpers = require('../helpers');
+var oauth = require('../oauth');
 
 /**
  * This controller logs in an existing user with Facebook OAuth.
@@ -36,7 +38,7 @@ module.exports = function (req, res) {
 
   if (!req.query.access_token) {
     logger.info('A user attempted to log in via Facebook OAuth without specifying an OAuth token.');
-    return res.status(400).json({ message: 'access_token parameter required.' });
+    return oauth.errorResponder(req, res, { message: 'access_token parameter required.' });
   }
 
   var userData = {
@@ -49,13 +51,13 @@ module.exports = function (req, res) {
   application.getAccount(userData, function (err, resp) {
     if (err) {
       logger.info('During a Facebook OAuth login attempt, we were unable to fetch the user\'s social account from Stormpath.');
-      return res.status(err.status || 400).json(err);
+      return oauth.errorResponder(req, res, err);
     }
 
     helpers.expandAccount(req.app, resp.account, function (err, expandedAccount) {
       if (err) {
         logger.info('During a Facebook OAuth login attempt, we were unable to fetch the user\'s social account from Stormpath.');
-        return res.status(err.status || 400).json(err);
+        return oauth.errorResponder(req, res, err);
       }
 
       res.locals.user = expandedAccount;
@@ -64,7 +66,7 @@ module.exports = function (req, res) {
       helpers.createStormpathSession(req.user, req, res, function (err) {
         if (err) {
           logger.info('During a Facebook OAuth login attempt, we were unable to create a Stormpath session.');
-          return res.status(err.status || 400).json(err);
+          return oauth.errorResponder(req, res, err);
         }
 
         var nextUrl = url.parse(req.query.next || '').path || (resp.created ? config.web.register.nextUri : config.web.login.nextUri);
--- a/lib/controllers/google-login.js
+++ b/lib/controllers/google-login.js
@@ -1,7 +1,9 @@
 'use strict';
 
-var helpers = require('../helpers');
 var url = require('url');
+
+var helpers = require('../helpers');
+var oauth = require('../oauth');
 
 /**
  * This controller logs in an existing user with Google OAuth.
@@ -31,7 +33,7 @@ module.exports = function (req, res) {
 
   if (!req.query.code) {
     logger.info('A user attempted to log in via Google OAuth without specifying an OAuth token.');
-    return res.status(400).json({ message: 'code parameter required.' });
+    return oauth.errorResponder(req, res, { message: 'code parameter required.' });
   }
 
   var userData = {
@@ -44,13 +46,13 @@ module.exports = function (req, res) {
   application.getAccount(userData, function (err, resp) {
     if (err) {
       logger.info('During a Google OAuth login attempt, we were unable to fetch the user\'s social account from Stormpath.');
-      return res.status(err.status || 400).json(err);
+      return oauth.errorResponder(req, res, err);
     }
 
     helpers.expandAccount(req.app, resp.account, function (err, expandedAccount) {
       if (err) {
         logger.info('During a Google OAuth login attempt, we were unable to fetch the user\'s social account from Stormpath.');
-        return res.status(err.status || 400).json(err);
+        return oauth.errorResponder(req, res, err);
       }
 
       res.locals.user = expandedAccount;
@@ -59,7 +61,7 @@ module.exports = function (req, res) {
       helpers.createStormpathSession(req.user, req, res, function (err) {
         if (err) {
           logger.info('During a Google OAuth login attempt, we were unable to create a Stormpath session.');
-          return res.status(err.status || 400).json(err);
+          return oauth.errorResponder(req, res, err);
         }
 
         var nextUrl = url.parse(req.query.next || '').path || (resp.created ? config.web.register.nextUri : config.web.login.nextUri);
--- a/lib/controllers/google-login.js
+++ b/lib/controllers/google-login.js
@@ -33,7 +33,7 @@ module.exports = function (req, res) {
 
   if (!req.query.code) {
     logger.info('A user attempted to log in via Google OAuth without specifying an OAuth token.');
-    return oauth.errorResponder(req, res, { message: 'code parameter required.' });
+    return oauth.errorResponder(req, res, new Error('code parameter required.'));
   }
 
   var userData = {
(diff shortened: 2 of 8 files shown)


lib/oauth/error-responder.js
```diff
@@ -0,0 +1,62 @@
...
+    var nextUri = url.parse(req.query.next || '').path;
+    var formActionUri = config.web.login.uri + (nextUri ? ('?next=' + nextUri) : '');
+    var oauthStateToken = resolveStateToken(req, res);
+
+    var hasSocialProviders = _.some(config.socialProviders, function (socialProvider) {
+      return socialProvider.enabled;
+    });
+
+    if (err.code === 7201) {
+      // Stormpath is unable to create or update the account because the
```
typerandom: I'd place this comment above `if (err.code === 7201) {`. Since it's related to that line.

lib/views/linkedin_login_form.jade
```diff
@@ -15,7 +15,9 @@ script(type='text/javascript').
           serializedQueryString += '&';
         }
 
-        serializedQueryString += key + '=' + encodeURIComponent(value);
+        if (key !== 'access_token') {
```
typerandom: I'd have this for loop just continue, instead of wrapping `serializedQueryString += key + '=' + encodeURIComponent(value);`. Also add a comment to why it is neccessary to explicitly strip this query parameter from the URL.

E.g.

```javascript
// Because the access token isn't bla bla bla...
if (key === 'access_token') {
  continue;
}

serializedQueryString += key + '=' + encodeURIComponent(value);
```

lib/oauth/error-responder.js
```diff
@@ -0,0 +1,62 @@
...
+ * @param {Object} err - The error to handle.
+ */
+function oauthErrorResponder(req, res, err) {
+  var accepts = req.accepts(['html', 'json']);
+  var config = req.app.get('stormpathConfig');
+
+  function renderLoginFormWithError(err) {
+    var view = req.app.get('stormpathConfig').web.login.view;
+    var nextUri = url.parse(req.query.next || '').path;
+    var formActionUri = config.web.login.uri + (nextUri ? ('?next=' + nextUri) : '');
```
typerandom: What happens if `login.uri` contains a query string? E.g. if it is `?page=login`. Also, you need to escape `nextUri`.

lib/oauth/error-responder.js
```diff
@@ -0,0 +1,62 @@
...
+ * @param {Object} req - The http request.
+ * @param {Object} res - The http response.
+ * @param {Object} err - The error to handle.
+ */
+function oauthErrorResponder(req, res, err) {
+  var accepts = req.accepts(['html', 'json']);
+  var config = req.app.get('stormpathConfig');
+
+  function renderLoginFormWithError(err) {
+    var view = req.app.get('stormpathConfig').web.login.view;
```
typerandom: It's nicer for readability to split these lines up.

```javascript
var stormpathConfig = req.app.get('stormpathConfig');
var view = stormpathConfig.web.login.view;
```

lib/oauth/error-responder.js
```diff
@@ -0,0 +1,62 @@
...
+
+var _ = require('lodash');
+var url = require('url');
+
+var helpers = require('../helpers');
+var forms = require('../forms');
+var resolveStateToken = require('./common').resolveStateToken;
+
+/**
+ * oauthErrorResponder takes an error object and respondes
```
typerandom: Remove `oauthErrorResponder` and capitalize `takes`. Also `respondes` should be `responds`.

lib/oauth/error-responder.js
```diff
@@ -0,0 +1,62 @@
+'use strict';
+
+var _ = require('lodash');
+var url = require('url');
+
+var helpers = require('../helpers');
+var forms = require('../forms');
+var resolveStateToken = require('./common').resolveStateToken;
```
typerandom: For readability, split this into two separate lines.

```javascript
var common =  require('./common');
var resolveStateToken = common.resolveStateToken;
```

lib/oauth/error-responder.js
```diff
@@ -0,0 +1,62 @@
...
+      formActionUri: formActionUri,
+      oauthStateToken: oauthStateToken,
+      hasSocialProviders: hasSocialProviders,
+      error: err.userMessage || err.message
+    };
+
+    helpers.render(req, res, view, options);
+  }
+
+  switch (accepts) {
```
typerandom: This seems overly engineered. The two cases to handle should only be HTML and JSON. So the likelihood of having to add support for some other format seems low. I'd rather just have this as a if, also, removes nesting.

```javascript
if (accepts === 'json') {
  return res.status(err.status || 400).json(err);
}

return renderLoginFormWithError(err);
```

lib/views/facebook_login_form.jade
```diff
@@ -16,6 +16,8 @@ script(type='text/javascript').
         var queryStr = window.location.search.replace('?', '');
         // TODO make dynamic
         if (queryStr) {
+          // Remove existing access token from query string.
```
typerandom: Same as my previous comment. Why do we want to remove this? Please add a comment describing why.

lib/controllers/facebook-login.js
```diff
@@ -36,7 +38,7 @@ module.exports = function (req, res) {
 
   if (!req.query.access_token) {
     logger.info('A user attempted to log in via Facebook OAuth without specifying an OAuth token.');
-    return res.status(400).json({ message: 'access_token parameter required.' });
+    return oauth.errorResponder(req, res, { message: 'access_token parameter required.' });
```
typerandom: To be consistent, maybe always pass an error object to `oauth.errorResponder()`? I.e.

```javascript
return oauth.errorResponder(req, res, new Error('access_token parameter required.'));
```

lib/controllers/facebook-login.js
```diff
@@ -49,13 +51,13 @@ module.exports = function (req, res) {
   application.getAccount(userData, function (err, resp) {
     if (err) {
       logger.info('During a Facebook OAuth login attempt, we were unable to fetch the user\'s social account from Stormpath.');
-      return res.status(err.status || 400).json(err);
+      return oauth.errorResponder(req, res, err);
```
typerandom: Same as previous comment.

lib/controllers/facebook-login.js
```diff
@@ -49,13 +51,13 @@ module.exports = function (req, res) {
...
     }
 
     helpers.expandAccount(req.app, resp.account, function (err, expandedAccount) {
       if (err) {
         logger.info('During a Facebook OAuth login attempt, we were unable to fetch the user\'s social account from Stormpath.');
-        return res.status(err.status || 400).json(err);
+        return oauth.errorResponder(req, res, err);
```
typerandom: Same as previous comment.

lib/controllers/google-login.js
```diff
@@ -31,7 +33,7 @@ module.exports = function (req, res) {
 
   if (!req.query.code) {
     logger.info('A user attempted to log in via Google OAuth without specifying an OAuth token.');
-    return res.status(400).json({ message: 'code parameter required.' });
+    return oauth.errorResponder(req, res, { message: 'code parameter required.' });
```
typerandom: Same as previous comment. Pass an `Error` object instead.

lib/views/facebook_login_form.jade
timothyej: I agree with you but don't think it's in the scope of this PR :) But I've added the comment and we can refactor this later (there's even a TODO comment to do so ;)).

lib/oauth/error-responder.js
timothyej: I was deciding between these two and was a bit sceptical myself. I've updated it to your suggestion now.

lib/oauth/error-responder.js
timothyej: Decided to just use `common.resolveStateToken()` where it's being used.

lib/oauth/error-responder.js
timothyej: Since this line is just copied from the login controller, I think that we should keep the same behaviour and try to refactor them at once (even better is to use the same code, and not copy). But this is too much work for the scope of this PR.

However, I did add the `encodeURIComponent`.

timothyej:
@typerandom Thanks for the review, I've addressed all of your comments in new commits :+1: Please review the code again, and also make sure to test the actual functionality (please note that LinkedIn might not work: #305). Also see my notes under *Discussion*.

robertjd:
Thanks @timothyej this looks great!  Yes, I think the error should be shown on the login form because that is the only place we render the social buttons, so it makes sense to place them there.

I added a commit, where I made a few small error rendering fixes. See https://github.com/stormpath/express-stormpath/commit/0ac9328e3f07c0053425745782fe309a1b3df63f

Issue #307, opened 2016-02-09
Implement JSON view model for GET on /register


timothyej:
Implement JSON view model for GET on `/register` according to the [framework spec](https://github.com/stormpath/stormpath-framework-spec):

> ## View Models
> 
> The login an registration views are dynamic.  The registration form has
> configurable fields, and both login and registration need to dynamically render
> buttons for any providers (Google, Facebook, SAML) that are mapped to the
> specified Stormpath application.
> 
> For that reason, we support JSON GET requests of the `/login` and `/register`
> views.  The response is a view model that the client can use to render the
> form appropriately.
> 
> Please see [login](https://github.com/stormpath/stormpath-framework-spec/blob/master/login.md) and [registration](https://github.com/stormpath/stormpath-framework-spec/blob/master/registration.md) for specific information on each
> model.

Part of #298.

timothyej:
As this is almost identical to #306, I will wait for #310 to be merged and then do a small refactor to make it work for `/register` as well.

PR #310, opened 2016-02-10
JSON view model for GET on /login

### LINKED ISSUE #306: Implement JSON view model for GET on /login
Implement JSON view model for GET on `/login` according to the [framework spec](https://github.com/stormpath/stormpath-framework-spec):

> ## View Models
> 
> The login an registration views are dynamic.  The registration form has
> configurable fields, and both login and registration need to dynamically render
> buttons for any providers (Google, Facebook, SAML) that are mapped to the
> specified Stormpath application.
> 
> For that reason, we support JSON GET requests of the `/login` and `/register`
> views.  The response is a view model that the client can use to render the
> form appropriately.
> 
> Please see [login](https://github.com/stormpath/stormpath-framework-spec/blob/master/login.md) and [registration](https://github.com/stormpath/stormpath-framework-spec/blob/master/registration.md) for specific information on each
> model.

Part of #298.

timothyej:
Fixes #306.

This PR implements a JSON view model for `GET` on `/login` according to the [spec](https://github.com/stormpath/stormpath-framework-spec/blob/master/login.md#login-view-model).

### Notes

* This PR is going to be merged into `framework-spec-upgrade`.

* I'm not sure if the account stores should only include SAML and social providers, or just all of them. If so, how do I filter them out?

* Added some unit tests (all methods are not tested):

  ```bash
helpers/get-login-view-model.js
  getFormFields(config)
    ✓ should return a non-empty array
    ✓ should only return fields with `enabled` set to `true`
    ✓ should remove the `enabled` property from returned fields
    ✓ should return the fields in order of `config.web.login.fieldOrder`
  getAccountStoreModel(accountStore)
    ✓ should return an object
    return object
      ✓ should not be the accountStore object
      ✓ should have href of accountStore
      ✓ should have name of accountStore
      .provider
        ✓ should have href of accountStore.provider
        ✓ should have callbackUri of accountStore.provider
        ✓ should have clientId of accountStore.provider
        ✓ should not have a clientSecret property
```

### How to verify

1. Checkout this branch
4. Make a `GET` request to `/login` with the `Accept` header set to `application/json`
5. Verify that you get a JSON model back according to the [spec](https://github.com/stormpath/stormpath-framework-spec/blob/master/login.md#login-view-model)
6. Try to edit the config (`web.login.fields` and `web.login.fieldOrder`) and verify that the fields and its order are correct
7. Verify the content of `accountStores` and see my notes about it

### DIFF
--- a/lib/controllers/login.js
+++ b/lib/controllers/login.js
@@ -10,7 +10,7 @@ var forms = require('../forms');
 var helpers = require('../helpers');
 var oauth = require('../oauth');
 var loginWithOAuthProvider = require('../helpers/login-with-oauth-provider');
-var getLoginViewModel = require('../helpers/get-login-view-model');
+var getLoginViewModel = require('../helpers/get-login-view-model').default;
 
 /**
  * This controller logs in an existing user.  If there are any errors, an
--- a/lib/controllers/login.js
+++ b/lib/controllers/login.js
@@ -10,6 +10,7 @@ var forms = require('../forms');
 var helpers = require('../helpers');
 var oauth = require('../oauth');
 var loginWithOAuthProvider = require('../helpers/login-with-oauth-provider');
+var getLoginViewModel = require('../helpers/get-login-view-model');
 
 /**
  * This controller logs in an existing user.  If there are any errors, an
@@ -50,6 +51,20 @@ module.exports = function (req, res) {
     });
 
     helpers.render(req, res, view, options);
+  }
+
+  // If a GET request is made with the Accept header as json,
+  // respond with the view model for this form.
+  // TODO: When `stormpath.web.produces` is implemented,
+  // this `if` must check that it is set to `application/json`.
+  if (req.method === 'GET' && accepts === 'json') {
+    return getLoginViewModel(config, function (err, viewModel) {
+      if (err) {
+        return res.status(400).json({ error: err.userMessage || err.message });
+      }
+
+      res.status(200).json(viewModel);
+    });
   }
 
   if (req.user && config.web.login.enabled) {
--- a/lib/controllers/login.js
+++ b/lib/controllers/login.js
@@ -10,7 +10,6 @@ var forms = require('../forms');
 var helpers = require('../helpers');
 var oauth = require('../oauth');
 var loginWithOAuthProvider = require('../helpers/login-with-oauth-provider');
-var getLoginViewModel = require('../helpers/get-login-view-model').default;
 
 /**
  * This controller logs in an existing user.  If there are any errors, an
@@ -58,7 +57,7 @@ module.exports = function (req, res) {
   // TODO: When `stormpath.web.produces` is implemented,
   // this `if` must check that it is set to `application/json`.
   if (req.method === 'GET' && accepts === 'json') {
-    return getLoginViewModel(config, function (err, viewModel) {
+    return helpers.getLoginViewModel(config, function (err, viewModel) {
       if (err) {
         return res.status(400).json({ error: err.userMessage || err.message });
       }
--- a/lib/helpers/get-login-view-model.js
+++ b/lib/helpers/get-login-view-model.js
@@ -158,10 +158,12 @@ function getLoginViewModel(config, callback) {
   });
 }
 
-module.exports = getLoginViewModel;
+module.exports = {
+  default: getLoginViewModel,
 
-// Expose private functions for testing.
-getLoginViewModel.prototype.getFormFields = getFormFields;
-getLoginViewModel.prototype.getAccountStores = getAccountStores;
-getLoginViewModel.prototype.getAccountStoreModel = getAccountStoreModel;
-getLoginViewModel.prototype.getProviders = getProviders;
+  // Expose private functions for testing.
+  getFormFields: getFormFields,
+  getAccountStores: getAccountStores,
+  getAccountStoreModel: getAccountStoreModel,
+  getProviders: getProviders
+};
--- a/lib/helpers/get-login-view-model.js
+++ b/lib/helpers/get-login-view-model.js
@@ -3,6 +3,8 @@
 var _ = require('lodash');
 var async = require('async');
 
+var cachedViewModel;
+
 /**
  * Returns list of all enabled fields, as defined by `stormpath.web.login.fields`,
  * and ordered by `stormpath.web.login.fieldOrder`.
@@ -131,6 +133,10 @@ function getProviders(config, callback) {
  * @param {Function} callback - Callback function (error, model).
  */
 function getLoginViewModel(config, callback) {
+  if (cachedViewModel) {
+    return callback(null, cachedViewModel);
+  }
+
   var fields = getFormFields(config);
 
   var model = {
@@ -146,6 +152,7 @@ function getLoginViewModel(config, callback) {
     }
 
     model.accountStores = providers;
+    cachedViewModel = model;
 
     callback(null, model);
   });
--- a/lib/helpers/get-login-view-model.js
+++ b/lib/helpers/get-login-view-model.js
@@ -4,6 +4,37 @@ var _ = require('lodash');
 var async = require('async');
 
 var cachedViewModel;
+var cacheTimestamp = 0;
+
+/**
+ * Returns the cached model if it's not older than `ttl`.
+ * Else it returns null.
+ *
+ * @method
+ *
+ * @param {number} ttl - The time-to-live value for the cached object.
+ */
+function getCachedModel(ttl) {
+  var age = (new Date().getTime() - cacheTimestamp) / 1000; // In seconds.
+
+  if (cachedViewModel && age <= ttl) {
+    return cachedViewModel;
+  }
+
+  return null;
+}
+
+/**
+ * Saves the model in "cache" together with a timestamp.
+ *
+ * @method
+ *
+ * @param {Object} model - The model to cache.
+ */
+function setCachedModel(model) {
+  cachedViewModel = model;
+  cacheTimestamp = new Date().getTime();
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by robertjd** [comment]
Hey @timothyej , I just gave this a quick look (I haven't pulled the branch yet).  RE which account stores to include: you want to include any account store where the provider ID is not 'stormpath'.

I see that you're querying for the account stores, every time you want to render this view model.  That is going to be expensive because we don't cache collections.  This view model should be cached.

File: lib/helpers/get-login-view-model.js [hunk 1]
```diff
@@ -0,0 +1,160 @@
...
+
+    model.accountStores = providers;
+
+    callback(null, model);
+  });
+}
+
+module.exports = getLoginViewModel;
+
+// Expose private functions for testing.
```
typerandom: I think this looks weird. If you want to test these, I'd rather have you export them. I.e.

```javascript
module.exports = {
  default: getLoginViewModel,
  getFormFields: getFormFields,
  getAccountStores: getAccountStores,
  getAccountStoreModel: getAccountStoreModel,
  getProviders: getProviders
};
```

Then from `helpers.js`:

```javascript
getLoginViewModel: require('./get-login-view-model').default
```

Then in your tests, simply just include `lib/helpers/get-login-view-model.js` directly. Or if they are generic enough, separate them to smaller helper utilities.

lib/controllers/login.js
```diff
@@ -10,6 +10,7 @@ var forms = require('../forms');
 var helpers = require('../helpers');
 var oauth = require('../oauth');
 var loginWithOAuthProvider = require('../helpers/login-with-oauth-provider');
+var getLoginViewModel = require('../helpers/get-login-view-model');
```
typerandom: Since `helpers` is already included, if `get-login-view-model` is in `helpers/index.js` you don't have to include it again.

lib/helpers/get-login-view-model.js
timothyej: This is not a solutions I'm entirely happy with either, but it looks better than mine so I went with this (bda7574dbddc89f97812a13c8b82fc62d0787964).

lib/controllers/login.js
timothyej: I'm not too happy about this *helper* file, and as you said yourself, this code should not even belong here. But in lack of a better structure at the moment, your suggestion is better and I fixed it (1923c19234ee0928e890f526f4be422e37fcaf73) :)

timothyej:
Thanks @robertjd! I've now implemented a very basic cache that's being cleared based on the default TTL from config. I've also filtered out all non-stormpath providers. Please have a look :)

lib/helpers/get-login-view-model.js
```diff
@@ -0,0 +1,176 @@
...
+    setTimeout(function () {
+      cachedViewModel = null;
+    }, cacheTtl * 1000);
+
+    callback(null, model);
+  });
+}
+
+module.exports = {
+  default: getLoginViewModel,
```
robertjd: Does assigning a value to module.exports.default actually do anything in a non-ES6 environment?

lib/helpers/get-login-view-model.js
```diff
@@ -0,0 +1,176 @@
...
+      return callback(err);
+    }
+
+    model.accountStores = providers;
+    cachedViewModel = model;
+
+    // Set timer to clear cache.
+    setTimeout(function () {
+      cachedViewModel = null;
+    }, cacheTtl * 1000);
```
robertjd: `setTimeout` and `setInterval` shouldn't be used for this, as they're unreliable.  Instead, store the time that the object was created and compare now against that time when you need to know if the object is stale.

lib/helpers/get-login-view-model.js
```diff
@@ -0,0 +1,176 @@
...
+  var options = { expand: 'accountStore' };
+
+  // Get account store mappings.
+  application.getAccountStoreMappings(options, function (err, accountStoreMappings) {
+    if (err) {
+      return callback(err);
+    }
+
+    // Get account stores.
+    async.map(accountStoreMappings.items, function (accountStoreMapping, done) {
```
robertjd: Why not use `accountStoreMappings.map` ?

robertjd:
Thanks @timothyej !  Please see inline comments for some additional feedback.  Looking good otherwise.

lib/helpers/get-login-view-model.js
timothyej: Nope, it was just a preference from Robin to do so (see our discussion above (the first comment)). Now I see that some files exports the private methods with a preceding `_`.

lib/helpers/get-login-view-model.js
timothyej: Got it! I've addressed this in commit 2ea060357c18556fc7b48127521579b26cd69677.

lib/helpers/get-login-view-model.js
timothyej: I forgot about this one, thanks! (7ae899cd2c0aeda8abbe9fadfe6eb053a8e96d6c)

timothyej:
@robertjd, thanks, I've addressed your comments :+1:

lib/helpers/get-login-view-model.js
```diff
@@ -0,0 +1,203 @@
...
+/**
+ * Returns the cached model if it's not older than `ttl`.
+ * Else it returns null.
+ *
+ * @method
+ *
+ * @param {number} ttl - The time-to-live value for the cached object.
+ */
+function getCachedModel(ttl) {
+  var age = (new Date().getTime() - cacheTimestamp) / 1000; // In seconds.
```
robertjd: Converting to seconds isn't really necessary.

lib/helpers/get-login-view-model.js
```diff
@@ -0,0 +1,203 @@
...
+ * <https://github.com/stormpath/stormpath-framework-spec/blob/master/login.md#login-view-model>
+ *
+ * @method
+ *
+ * @param {Object} config - The Stormpath configuration.
+ * @param {Function} callback - Callback function (error, model).
+ */
+function getLoginViewModel(config, callback) {
+  var cacheTtl = config.client.cacheManager.defaultTtl;
+  var cached = getCachedModel(cacheTtl);
```
robertjd: Let's call this `cachedModel`, for clarity.

lib/helpers/get-login-view-model.js
```diff
@@ -0,0 +1,203 @@
...
+
+  if (cached) {
+    return callback(null, cached);
```
robertjd: Please use process.nextTick here.  We want to follow the paradigm that asynchronous API's should actually be asynchronous.

robertjd:
Hey @timothyej , please see three new inline comments (nextTick, milliseconds, variable naming) for some simple fixes.  Other than those, I’m happy with this PR and it can be merged.

Looking at the conversation, it seems that you’re trying to have a single “public” export function, that is meant to be used publicly, but you also want to make it easy to test the internal workings of this function.

Breaking it up into smaller functions and exporting them all is certainly one way to go, but it can look/feel messy from the viewpoint of the public API.  In this situation, you have to use the `_methodName()` convention, or use comments to declare what is internal and what is public - so that developers don’t depend on things that might change in the future.

If I may suggest an alternative: this entire operation can be treated as a single asynchronous task.  In doing so, you have to ask: what inputs are needed for this task?  In this particular situation, those inputs are the application and framework configuration.  As such, the single public interface would look like this:

```
module.exports = function getLoginViewModel(application, config, callback){
  var viewModel;
  // .. all the things
  callback(err,viewModel);
}
```

In order to test this function, simply create mocks (for the application and config) and then assert that the callback value is appropriate, given the inputs.  To test the cache behavior, you should spy on your mock methods that supply the mock account store values, and assert that they are called when they should, given the input cache configuration.  This is a test that would happen over time, and *would* require a setTimeout to allow some time to pass (please don’t use time testing libraries that muck with the JS time objects).  In short: you’re testing the side effects of the public interface, and not the internal implementation details.  This is totally fine, and you’ll still get to full code coverage with this approach.  It’s also beautiful because it allows you to refactor your internal implementation, while still trusting that your public interface is doing the right thing.

I hope these ideas help!

lib/helpers/get-login-view-model.js
timothyej: :+1:

lib/helpers/get-login-view-model.js
timothyej: TIL :+1:

lib/helpers/get-login-view-model.js
timothyej: The `ttl` is in seconds, so the conversion must happen here or in the `if` two lines down. I prefer to normalize the age here so we have both `ttl` and `age` in seconds. Or am I missing something?

timothyej:
Thanks @robertjd for the comments, I've addressed them all :)

And regarding the tests:

**I really like your idea because:**

* It only tests the API and not the implementation (refactors is easier for example)
* We don't have to deal with the problem of exposing private methods for testing

**My concerns about this approach are:**

* The tests *might* be harder to structure (as we have to test every case of the method from the surface)
* The tests *might* not reveal what's actually wrong when something is failing

But these are just theories and I'm looking forward to go with your approach the next time I write tests. We can continue this discussion then :)

timothyej:
I'm going ahead and merging this as I need this to proceed with #307. Any concerns from this PR can be addressed in the (future) PR for #307.

lib/helpers/get-login-view-model.js
robertjd: Right, that makes sense.  Thanks.

Issue #309, opened 2016-02-10
Implement produces option for /register


typerandom:
Implement the produces option for `/register` according to the [framework spec](https://github.com/stormpath/stormpath-framework-spec/blob/master/registration.md#feature-description):

>##### GET requests must respond in one of the following ways:
>
>Serve a default HTML page with a registration form, if the request type is Accept: text/html and text/html is defined in stormpath.web.produces.
>
>Serve the developer's Single Page Application, if stormpath.web.spa.enabled is true and text/html is defined in stormpath.web.produces.
>
>Serve the registration view model if the request type is Accept: application/json and application/json is defined in stormpath.web.produces.

Part solution of #298.

PR #313, opened 2016-02-11
Add produces option for /login

### LINKED ISSUE #308: Implement produces option for /login
Implement the produces option for `/login` according to the [framework spec](https://github.com/stormpath/stormpath-framework-spec/blob/master/login.md#feature-description):

>##### GET requests must:
>
>Serve a default HTML page with a login form, if the request type is Accept: text/html and text/html is defined in stormpath.web.produces.
>
>Serve the developer's Single Page Application, if stormpath.web.spa.enabled is true and text/html is defined in stormpath.web.produces.
>
>Serve the login view model if the request type is Accept: application/json and application/json is defined in stormpath.web.produces.

Part solution of #298.

rorheden:
Implements the produces option for /login according to [the framework spec](https://github.com/stormpath/stormpath-framework-spec/blob/master/login.md#feature-description).
#### How to verify
1. Checkout this branch.
4. Start the server.
5. Navigate to `http://localhost:3000/login`.
6. Login with an existing user.
7. Open up the terminal and execute `$ http POST localhost:3000/login username=abc password=123 --json`.
8. Verify that you get a `400 Bad Request` with a `Content-Type: application/json; charset=utf-8` and `{"error":"Invalid username or password."}` body.
9. Redo step 7, but this time change the username and password to a real account.
10. Verify that you get a `200 OK` response with two `Set-Cookie` headers. One for setting the `access_token` cookie, and the other one for setting `refresh_token`.
11. Now modify your configuration, and change `web.produces` to `['text/html', '']` (hack because extender currently merges arrays... which needs to be fixed, empty string overwrites second array item.).
12. Verify that step 5-6 works, but that step 7 responds with a `404 Not Found` status.
13. Modify the `web.produces` config to `['application/json', '']`.
14. Verify that step 5 results in a `404 Not Found` and that steps 7-10 work as previous.
15. Modify the `web.produces` config again to `['', '']`.
16. Verify that both steps 5 and 7 result in a ´404 Not Found`.

Fixes #308.


timothyej:
Dayum! Me personally would have waited with a refactor of this size until after the *framework spec upgrade*. But it looks nice, I'm looking forward to review it later :)

typerandom:
@timothyej It was necessary in order to implement this easily. Since I'll also have to add the same functionality for `/register` (#309). And also makes fixing the SPA options issue a lot easier.

timothyej:
I'm okay with it, as long as we make sure to merge these PRs before starting on the other things, and do proper reviews :+1:

typerandom:
@timothyej Yeah, that is also the thought.

lib/controllers/login.js
```diff
@@ -21,121 +19,132 @@ var loginWithOAuthProvider = require('../helpers/login-with-oauth-provider');
  * @param {Object} req - The http request.
  * @param {Object} res - The http response.
  */
-module.exports = function (req, res) {
-  var accepts = req.accepts(['html', 'json']);
+module.exports = function (req, res, next) {
```
timothyej: Please document the new `next` parameter in the JS docs above.

typerandom: Thanks! Yeah, I'll fix that.

lib/controllers/login.js
```diff
@@ -21,121 +19,132 @@ var loginWithOAuthProvider = require('../helpers/login-with-oauth-provider');
...
-    return res.redirect(302, nextUrl);
+        callback();
+      });
+    });
   }
 
-  res.locals.status = req.query.status;
+  helpers.handleResponse(req, res, next, {
+    'application/json': function () {
+      var self = this;
```
timothyej: I'm not a big fan of `self` :P I would rather do something like this:

```javascript
var accept = this.accept;
var reject = this.reject;
```

And then use `accept` and `reject` in the other scope further down.

lib/controllers/login.js
```diff
@@ -21,121 +19,132 @@ var loginWithOAuthProvider = require('../helpers/login-with-oauth-provider');
...
+        self.accept();
       });
-    });
-  } else if (accepts === 'html') {
-    if (config.web.spaRoot) {
-      return res.sendFile(config.web.spaRoot);
-    }
+    },
+    'text/html': function () {
+      var self = this;
```
timothyej: Same here with `self`, if possible :)

lib/controllers/login.js
typerandom: Yeah, I think that is bloating it a lot. Which I don't really like.

lib/controllers/login.js
typerandom: Yeah, same as before. I think that is unnecessary bloat.

lib/controllers/login.js
timothyej: Okay, but how about using `.bind()`, which is pretty much there to solve this issue?

```javascript
authenticate(req.body, function (err) {
  if (err) {
    return this.reject(err);
  }

  this.accept();
}.bind(this));
```

Only downside with this is that it is easy to not see that `.bind()` and therefore not know that the function is bound to another scope.

typerandom: Yeah, same as before. I think that is necessary bloat. I don't really feel that this solves any issue. I don't like `self` that much myself. But before there is a really good solution to it (as ES6 arrow functions), I think that's the best we've got.

timothyej: I don't see how this adds bloat, as it actually reduces your code by 1 line. And `.bind()` is a valid solution to this problem, as `self` is more like a hack.

typerandom: Except that it needs to be added on every method invoked, and it's easy to miss that. Which makes it prone to errors. Self is a hack, but it's very easy to use and understand and hard to get wrong.

timothyej: Okay, let's just hold out until we can use the ES6 arrow functions then :)

typerandom: Yep, that is my idea:) That will really clean up a lot of this crap.

typerandom:
Please review this again. For some reason the tests are failing when running on Travis, but pass when I run it locally. Everything should be pushed. Please test and see if you can reproduce the same errors as shown on Travis.

typerandom:
Please review.

test/helpers/test-handle-response.js
```diff
@@ -0,0 +1,455 @@
...
+
+          it('should have a req property', function () {
+            assert('req' in handlerScope);
+          });
+
+          it('should have a res property', function () {
+            assert('res' in handlerScope);
+          });
+
+          it('should have a accept() method', function () {
```
timothyej: `a` => `an`

test/helpers/test-handle-response.js
```diff
@@ -0,0 +1,455 @@
...
+
+          it('should have a reject() method', function () {
+            assert(typeof handlerScope.reject === 'function');
+          });
+
+          it('should have a pass() method', function () {
+            assert(typeof handlerScope.pass === 'function');
+          });
+
+          it('should have a isGetRequest() method', function () {
```
timothyej: `a` => `an`

test/helpers/test-handle-response.js
```diff
@@ -0,0 +1,455 @@
...
+            assert(typeof handlerScope.isGetRequest === 'function');
+          });
+
+          it('should have a isPutRequest() method', function () {
```
timothyej: `a` => `an`

test/helpers/test-handle-response.js
```diff
@@ -0,0 +1,455 @@
...
+            assert(typeof handlerScope.isPutRequest === 'function');
+          });
+
+          it('should have a isPostRequest() method', function () {
```
timothyej: `a` => `an`

test/helpers/test-handle-response.js
```diff
@@ -0,0 +1,455 @@
...
+            assert(typeof handlerScope.isPostRequest === 'function');
+          });
+
+          it('should have a isDeleteRequest() method', function () {
```
timothyej: `a` => `an`

test/helpers/test-handle-response.js
```diff
@@ -0,0 +1,455 @@
...
+          assert(!spies.responseStatus.called);
+        });
+
+        it('shouldn\'t call res.end()', function () {
+          assert(!spies.responseEnd.called);
+        });
+      });
+    });
+
+    describe('a application/json handler', function () {
```
timothyej: `a` => `an`

test/helpers/test-handle-response.js
```diff
@@ -0,0 +1,455 @@
...
+      mocks = {};
+      mocks.config = { web: { produces: ['text/html'] } };
+      mocks.app = new MockApp(mocks.config);
+      mocks.response = new MockResponse();
+      mocks.request = new MockRequest('GET', mocks.app, 'text/html');
+
+      handleResponse(mocks.request, mocks.response, spies.next, {}, spies.callback);
+    });
+
+    it('is called', function () {
```
timothyej: What about `it('should be called')`?

typerandom: +1. Sounds a lot better. Going with that.

test/helpers/test-handle-response.js
typerandom: Thanks! Fixed!

test/helpers/test-handle-response.js
typerandom: Thanks! Fixed!

test/helpers/test-handle-response.js
typerandom: Thanks! Fixed!

test/helpers/test-handle-response.js
typerandom: Thanks! Fixed!

test/helpers/test-handle-response.js
typerandom: Thanks! Fixed!

test/helpers/test-handle-response.js
typerandom: Thanks! Fixed!

typerandom:
@timothyej Fixed the things you mentioned. Any other feedback? Can we merge this?

timothyej:
Not yet, will review a little bit more in a moment.

typerandom:
@timothyej Thanks! :)

robertjd:
Hey @typerandom , I’m also having a strong reaction to the refactoring that is going on here.  It appears that you’re re-inventing an HTTP framework, inside of Express.  This is not necessary, because Express provides functionality that you can re-use.  With these changes, it’s hard to follow our logic, and is problematic for our developers.  If they know Express, they should be able to understand our code with ease.  Our 1.x -> 2.x refactor was written with readability in mind, and we saw some great PRs as a result.  I want to preserve that ability for our developers.

Specifically, the `pass` and `accept` concepts are a surprise, and I have to dig into figure out what is going on.  When I dig in I find `next()` and a `callback()` being called at the same time.  To an async developer, this isn’t feeling good.

I need to push back on this approach, and request that you stick with Express conventions.  It’s also unclear to me if this will integrate well with #310?

From a feature-ready perspective, I’m trying to use this configuration:

```
app.use(stormpath.init(app, {
  website: true,
  produces: ['application/json'],
  debug: 'info, error'
}));
```

But I’m still able to get an HTML response from the /login route, even though I’ve omitted that from the produces option.  In this situation I would expect a 404 when I visit /login from the browser, when it asks for HTML.

Some more specific notes:

* Please revisit all JSON error handlers to ensure that they are following the new [Error Handling Guidelines](https://github.com/stormpath/stormpath-framework-spec/blob/master/error-handling.md)
* We need to accept either `login` or `username` as valid properties for login.  In 3.0 we will remove `username` in favor of `login`.

typerandom:
@robertjd 

>I’m also having a strong reaction to the refactoring that is going on here. It appears that you’re re-inventing an HTTP framework, inside of Express. This is not necessary, because Express provides functionality that you can re-use. With these changes, it’s hard to follow our logic, and is problematic for our developers. If they know Express, they should be able to understand our code with ease. Our 1.x -> 2.x refactor was written with readability in mind, and we saw some great PRs as a result. I want to preserve that ability for our developers.

I don't agree with that. It's not reinventing the HTTP framework in express. It's creating a generic wrapper for the way that we work with certain requests/responses. I also see it as more readable and less error-prone, since we have a lot less code to manage. And also since we enforce a certain way to work, it enforce consistency across all the requests/responses.

>Specifically, the pass and accept concepts are a surprise, and I have to dig into figure out what is going on. When I dig in I find next() and a callback() being called at the same time. To an async developer, this isn’t feeling good.

There's a big difference between `callback` and `next`. Callback is a callback function for when the `handleResponse` helper has finished its work, and `next` is a method used in express to tell it to continue processing its middleware pipeline. If you have a look in `login.js`, you will see that the callback isn't even being used since we're not interested in knowing the end result. If that adds confusion, we could even remove it. I just added it, since I thought it could help when debugging.

>From a feature-ready perspective, I’m trying to use this configuration:
>```
>app.use(stormpath.init(app, {
  website: true,
  produces: ['application/json'],
  debug: 'info, error'
}));
```

It might not be that obvious. But I mention this issue in the "How to verify" section (see step 11). The reason for `produces: ['application/json']` not working is due to an issue with the way that configuration is being extended. The extender lib does not overwrite the array, it tries to extend it. So in order to set the right configuration you need to set it to `produces: ['application/json', '']` (this will overwrite the second default array value).

**Comment by typerandom** [comment]
Also, I just noticed one thing. `handleResponse` is kind of wrong. It should be `handleRequest`, since we're actually handling a request, and not the response.

**Comment by robertjd** [comment]
Hey @typerandom , thanks for the clarifications.  RE the issues I had with the produces option: yes, that was my bad for reading the notes incorrectly.  Let’s get those array merging issues resolved before we make this 2.5 release.  RE my concerns over the HTTP abstraction, please follow the feedback from our chat earlier: to rename the generic request handler to something that indicates that it’s delegating based on content type (I do see value in that abstraction), and to have the content type handlers take the `(req, res, next)` form and use those existing Express APIs for “passing” on the request, or ending it with our own response.  If you want to make error handling more dry, please continue with the "helper" approach where we export/require a function that does the common task.  This kind of function composition is working really well for this library.  Thanks for chatting it out!

**Comment by timothyej** [comment]
@typerandom, let me know when you're done with the changes and I will do my final review then :) I love the tests btw :heart:

**Comment by typerandom** [comment]
@timothyej Thanks! Unfortunately I had to remove that now when I had to refactor it.

@robertjd @timothyej Please review again. Once reviewed and we're ok with the solution, I'll also add some tests.

File: lib/controllers/login.js [hunk 11]
```diff
@@ -4,152 +4,171 @@ var _ = require('lodash');
...
-      oauthStateToken: oauthStateToken,
-      hasSocialProviders: hasSocialProviders
-    });
+  credentials = credentials || {};
+  credentials.username = credentials.login || credentials.username || '';
+  credentials.password = credentials.password || '';
 
-    helpers.render(req, res, view, options);
+  if (!credentials.username || !credentials.password) {
+    return callback(new Error('Invalid username or password.'));
```
robertjd: We should let the authenticator return this error, so that we aren't duplicating this error in a second location.

typerandom: Yes, I agree. It was just a fix from the previous solution by defaulting to `uuid()`, which was pretty bad.

lib/controllers/login.js
```diff
@@ -4,152 +4,171 @@ var _ = require('lodash');
...
-          accessToken: providerData.accessToken,
-          code: providerData.code
-        }
-      };
+  helpers.handleContentTypeRequest(req, {
+    'application/json': function () {
+      switch (req.method) {
+        // Respond with the view model for this form.
+        // TODO: When `stormpath.web.produces` is implemented,
+        // this `if` must check that it is set to `application/json`.
```
robertjd: I think this comment block is old?

typerandom: Yeah, it was one that Timothy added that got thrown in during my rebase. I'll fix that.

lib/controllers/login.js
robertjd: Hrm, we can't pass username as an empty string to the SDK?

typerandom: Yeah, we can. But the API will not return a `Invalid username or password.` error, but something like `Username or password cannot be an empty string.`.

lib/controllers/login.js
```diff
@@ -4,152 +4,168 @@ var _ = require('lodash');
...
-      // Social Login
-      var options = {
-        providerData: {
-          providerId: providerData.providerId,
-          accessToken: providerData.accessToken,
-          code: providerData.code
-        }
-      };
+  helpers.handleContentTypeRequest(req, {
+    'application/json': function () {
```
timothyej: How about extracting these handler functions and giving them a name? Right now it's hard to get an overview of what types of responses can be made. Something like this maybe:

```javascript
helpers.handleContentTypeRequest(req, {
  'application/json': jsonHandler(req, res, next),
  'text/html': htmlHandler(req, res, next)
});
```

(`jsonHandler` and `htmlHandler` would return the handler methods.)

lib/controllers/login.js
```diff
@@ -4,152 +4,168 @@ var _ = require('lodash');
...
-    if (providerData) {
-      // Social Login
-      var options = {
-        providerData: {
-          providerId: providerData.providerId,
-          accessToken: providerData.accessToken,
-          code: providerData.code
-        }
-      };
+  helpers.handleContentTypeRequest(req, {
```
timothyej: > `handleContentTypeRequest`

The `Content-Type` header is used by the response. The request is using the `Accept` header. So the name `handleContentTypeRequest` sounds a bit wrong in that sense.

timothyej:
I've verified that this works as advertised :+1: But please see my comments.

lib/controllers/login.js
typerandom: My bad. What do you think about `handleAcceptRequest`?

lib/controllers/login.js
typerandom: I think _hard_ is exaggerated. But I'll look into it.

lib/controllers/login.js
robertjd: Thanks.  So the `Username or password cannot be null or empty` error is actually coming from the REST API.  I do prefer this error because it's more descriptive (it tells you that you are missing a field), albeit confusing now that we're calling this field `login` at the framework layer.  Blarg.  I'm opening a discussion [here](https://github.com/stormpath/stormpath-framework-spec/issues/36).  Let's leave this as-is for now and revisit per discussion output.  This doesn't hold up this PR.

lib/controllers/login.js
robertjd: > The Content-Type header is used by the response. 

@timothyej that's not correct, the `Content-Type` header declares the format of the incoming POST body - as such this is related to the request, not the response.

Your other concern is valid, `handleAcceptRequest` would be more correct as the function is branching based on the Accept header.

lib/controllers/login.js
robertjd: Hey guys, I appreciate the proactive refactoring.  However, these types of changes should happen outside of a feature sprint (the 2.5 milestone we are working on).  If we are going to introduce HTTP abstractions we need to design those as a team, so that everyone is on board.  Unfortunately there just isn’t time for this right now, as we need to get this 2.5 release out ASAP.  As such, please do not introduce any more HTTP abstractions as this time.  In this particular case, our previous approach of using the `req.accepts()` API, with simple if statements, is working perfectly fine.

robertjd:
Hey all, I'm going to merge this now.  Here is my final follow up:

* For now we'll leave that extra error handler for the null username case, see inline comments and [framework-spec/36](https://github.com/stormpath/stormpath-framework-spec/issues/36)
* Renamed handleContentTypeRequest -> handleAcceptRequest, per the discussion I saw.  However let's stop at this, please no more abstraction refactoring during milestone sprints (see extra message I left in hipchat, I don't know where the inline comment went on github).
* Thanks @typerandom for pulling out what we discussed, this looks much cleaner now.
* I added a fix in the login view model builder, it needed to exclude account stores that are organizations because they don't have a provider.  This was causing an exception.

Issue #316, opened 2016-02-12
Add support for stormpath.web.social configuration


timothyej:
Add support for [`stormpath.web.social`](https://github.com/stormpath/stormpath-framework-spec/blob/master/social.md#implementing-page-based-workflows) option.

>The library must implement provider callbacks for all social provider directories that are mapped to the specified Stormpath application. The callback URLs for each provider should take the form of:
>
> >`/<stormpath.web.social.[providerId].uri>`
>
> Where `providerId`  is the property that is found in the `provider` object of the Stormpath directory, e.g. `google` or `facebook`.

Part of #298.

timothyej:
Closed in favour of https://github.com/stormpath/stormpath-node-config/issues/32

PR #318, opened 2016-02-12
Add web.socialProviders.callbackRoot to the config

timothyej:
This PR only changes the default config to include the new config parameter `web.socialProviders.callbackRoot`. The actual functionality is handled in the node config (https://github.com/stormpath/stormpath-node-config/issues/32).

### Discussion

* The [spec](https://github.com/stormpath/stormpath-framework-spec/blob/master/social.md#implementing-page-based-workflows) says that the callback url should be built like this:

  > `/<stormpath.web.socialProviders.callbackRoot>/<providerId>`

  That would mean that `stormpath.web.socialProviders.callbackRoot` would not start with a `/`, and would instead be just `callbacks`.

  To be consistent with other paths/URIs in the config, I decided to include the `/` at the beginning, i.e. `/callbacks`. Whether what is correct or not, this should be corrected/clarified in the specs.

* The current social providers are located at `stormpath.socialProviders` (without `web`). Is this intentional or should this be changed as well?

### How to review

* Make sure that the config parameter looks correct and its default value
* Look at the *Discussion* before merging

### DIFF
--- a/lib/config.yml
+++ b/lib/config.yml
@@ -22,6 +22,8 @@ web:
     secure: null
     path: "/"
     domain: null
+  socialProviders:
+    callbackRoot: "/callbacks"
   # The registration feature will be automatically enabled by the existence of
   # a default account store for your application.
   register:


robertjd:
I've opened this PR on the spec, let's try to get some of this resolved before we move forward: https://github.com/stormpath/stormpath-framework-spec/pull/35

typerandom:
@timothyej Do you have any update on this one? Is it still pending on the framework spec issue?

typerandom:
@timothyej Bump. Any update here?

timothyej:
@typerandom Yes, please see Robert's comment. There will need to be changes here no matter what.

robertjd:
Closing, we updated the social configuration in the framework spec and this PR is no longer accurate.  I reopend #316

PR #323, opened 2016-02-16
JSON view model for GET on /register

### LINKED ISSUE #307: Implement JSON view model for GET on /register

timothyej:
Fixes #307.

This PR implements a JSON view model for `GET` on `/register` according to the [spec](https://github.com/stormpath/stormpath-framework-spec/blob/master/registration.md#registration-view-model).

This is done by refactoring the `helper/get-login-view-model.js` to be able to take a form for which it generates a view model for (f6c911b).

### Notes

* This PR is going to be merged into `framework-spec-upgrade`.

### How to verify

1. Checkout this branch
4. Make a `GET` request to `/login` with the `Accept` header set to `application/json`
5. Verify that you get a JSON model back according to the [spec](https://github.com/stormpath/stormpath-framework-spec/blob/master/registration.md#registration-view-model)
6. Try to edit the config (`web.register.fields` and `web.register.fieldOrder`) and verify that the fields and its order are correct
7. Verify the content of `accountStores`
8. Try the same for `/login`


robertjd:
@timothyej I added a fix that has broken this PR, sorry! :/ please see https://github.com/stormpath/express-stormpath/commit/a2d4f8f38bc3765851dd06911238ab7bafd19f75 in #313, I had to add a fix for the account store model builder.  Please bring that change over here too.

timothyej:
@robertjd No problem, I was expecting to solve merge conflicts :) They're now fixed, so please give it another review.

lib/helpers/get-form-view-model.js
```diff
@@ -12,13 +11,20 @@ var cacheTimestamp = 0;
...
  */
-function getCachedModel(ttl) {
-  var age = (new Date().getTime() - cacheTimestamp) / 1000; // In seconds.
+function getCachedModel(formName, ttl) {
+  var cacheItem = cache[formName];
 
-  if (cachedViewModel && age <= ttl) {
-    return cachedViewModel;
+  if (!cacheItem) {
+    return null;
```
typerandom: Minor nitpick. I'd rather return false here instead of null. Since `false` would indicate that there isn't anything.

robertjd: My habit, for situations where the return value is being used for a truthy test, is to use `return;`.  However I agree that `false` is more explicit and I like that.  If we're all in agreement I'd like to see an eslint rule that enforces no-null-returns, that should come as a separate PR after we clear this milestone.

robertjd:
Thanks @timothyej !  This PR looks great, +1 to merge.  Please see comments RE returning false instead of null, we should keep discussing but it doesn't need to block this merge.

PR #331, opened 2016-02-17
Add web.spa option support

### LINKED ISSUE #319: Implement SPA configuration options
Implement SPA configuration options according to the [spec](https://github.com/stormpath/stormpath-framework-spec/blob/master/single-page-apps.md#configuration).

```yml
stormpath:
  web:
    spa:
      enabled: true
      view: index  # or other value that indicates where the entry is
```

Part of #298.

typerandom:
Adds support for the web.spa option.

#### Before reviewing

This depends on #313 to be merged before being reviewed.

#### Discuss

What should the default `web.spa.view` value be? I've just set it to `index`, since that is [what the framework spec states](https://github.com/stormpath/stormpath-framework-spec/blob/master/web-config.yaml#L214). But doesn't really make any sense since there won't be any file named `index`.

One solution that I think makes sense would be to set it to `null`, then have a validation step when the configuration is being loaded, to assert that if `web.spa.enabled` is set to `true`, then a value for `web.spa.view` must also be provided.

#### How to verify

1. Checkout this branch.
2. Run `$ npm link`.
3. Checkout the [React example app](https://github.com/stormpath/stormpath-express-react-example).
4. Run `$ npm link express stormpath`.
5. Open up `server.js`.
6. Remove the old `web.spaRoot` option.
7. Set `web.spa.enabled` to `true` and `web.spa.view` to `path.join(__dirname, 'build/index.html')`.
8. Run `$ npm start`.
9. Navigate to [http://locahost:3000/login](http://locahost:3000/login).
10. Verify that the React login view (`build/index.html`) is shown.
11. Open up `server.js`and set `web.spa.enabled` to `false`.
12. Run `$ npm start`.
13. Navigate to [http://locahost:3000/login](http://locahost:3000/login).
14. Verify that the Express Stormpath login view is shown.

Fixes #319.

### DIFF
--- a/lib/helpers/handle-accept-request.js
+++ b/lib/helpers/handle-accept-request.js
@@ -10,12 +10,18 @@
  * @param {Object} handlers - Object where the key is the content type to handle.
  * @param {function} fallbackHandler - Handler to call when no content type was matched.
  */
-function handleAcceptRequest(req, handlers, fallbackHandler) {
+function handleAcceptRequest(req, res, handlers, fallbackHandler) {
   var config = req.app.get('stormpathConfig');
   var accepted = req.accepts(config.web.produces);
 
   if (!accepted || !(accepted in handlers)) {
     return fallbackHandler();
+  }
+
+  // If this is a text/html response, then check if we have SPA-mode enabled.
+  // If it is enabled, then return the view for it.
+  if (accepted === 'text/html' && config.web.spa.enabled) {
+    return res.sendFile(config.web.spa.view);
   }
 
   handlers[accepted]();
--- a/lib/helpers/handle-accept-request.js
+++ b/lib/helpers/handle-accept-request.js
@@ -1,14 +1,14 @@
 'use strict';
 
 /**
- * Handles a content type request.
+ * Handles a HTTP Accept request.
  *
  * @method
  * @private
  *
  * @param {Object} req - HTTP request.
- * @param {Object} handlers - Object where the key is the content type to handle.
- * @param {function} fallbackHandler - Handler to call when no content type was matched.
+ * @param {Object} handlers - Object where the key is the content type's available to accept.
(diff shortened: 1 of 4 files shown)


lib/helpers/handle-accept-request.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+
+/**
+ * Handles a HTTP Accept request.
```
timothyej: Typo: `a` => `an`

lib/helpers/handle-accept-request.js
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * @method
+ * @private
+ *
+ * @param {Object} req - HTTP request.
```
timothyej: Don't forget to add the `res` parameter here too :)

timothyej:
Except for my minor comments, this works just fine :+1:

lib/helpers/handle-accept-request.js
typerandom: Fixed:)

lib/helpers/handle-accept-request.js
typerandom: Fixed:) Thanks!

typerandom:
@timothyej Thanks! I have updated the PR according to your comments:)

robertjd:
+1, this looks great!  RE the default for the view, yeah I agree that it should be null with a config error.  I'm [pinging les](https://github.com/stormpath/stormpath-framework-spec/issues/39) to double check what the reasoning was, but I feel comfortable doing the null approach in our library, so please feel free to make that change here. Otherwise, ready to merge.

timothyej:
@typerandom can you check why the tests fails?

typerandom:
@timothyej Fixed :)

timothyej:
:+1:

PR #332, opened 2016-02-17
Add produces option for /register

### LINKED ISSUE #309: Implement produces option for /register

typerandom:
Implements the produces option for /register according to [the framework spec](https://github.com/stormpath/stormpath-framework-spec/blob/master/registration.md#feature-description).

#### How to verify

1. Checkout this branch.
4. Start the server.
5. Navigate to `http://localhost:3000/register`.
6. Sign up for a new account.
7. Open up the terminal and execute `$ http POST localhost:3000/register givenName=test surname=tester email=[your name]+123@stormpath.com password=Test12345_ --json`.
8. Verify that you get a `200 OK` response with the account data as a JSON body.
9. Modify your configuration and change `web.produces` to `['text/html', '']` (hack because extender currently merges arrays... which needs to be fixed, empty string overwrites second array item.).
10. Verify that step 5-6 works, but that step 7 responds with a `404 Not Found` status.
11. Modify the `web.produces` config to `['application/json', '']`.
12. Verify that step 5 results in a `404 Not Found` and that steps 7-8 work as previous.
13. Modify the `web.produces` config again to `['', '']`.
14. Verify that both steps 5 and 7 result in a ´404 Not Found`.

Fixes #309.

### DIFF
--- a/lib/controllers/register.js
+++ b/lib/controllers/register.js
@@ -125,7 +125,7 @@ module.exports = function (req, res, next) {
   var postRegistrationHandler = config.postRegistrationHandler;
   var view = config.web.register.view;
 
-  helpers.handleAcceptRequest(req, {
+  helpers.handleAcceptRequest(req, res, {
     'application/json': function () {
       switch (req.method) {
         case 'GET':
@@ -189,17 +189,10 @@ module.exports = function (req, res, next) {
           next();
       }
     },
-    // Handle incoming POST requests from an browser-like clients (something like
-    // chrome / firefox / etc.).
     'text/html': function () {
       switch (req.method) {
         // We should render the registration template.
         case 'GET':
-          if (config.web.spaRoot) {
-            return res.sendFile(config.web.spaRoot);
-          }
-
-          // Render the registration template.
           helpers.render(req, res, view, {
             form: helpers.sanitizeFormData(req.body, config),
             formModel: formModel
--- a/lib/controllers/register.js
+++ b/lib/controllers/register.js
@@ -114,9 +114,9 @@ function defaultJsonErrorResponder(err, res) {
  *
  * @param {Object} req - The http request.
  * @param {Object} res - The http response.
+ * @param {function} next - The next callback.
  */
-module.exports = function (req, res) {
-  var accepts = req.accepts(['html', 'json']);
+module.exports = function (req, res, next) {
   var application = req.app.get('stormpathApplication');
   var authenticator = new stormpath.OAuthPasswordGrantRequestAuthenticator(application);
   var config = req.app.get('stormpathConfig');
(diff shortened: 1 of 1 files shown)


lib/controllers/register.js
```diff
@@ -125,159 +125,174 @@ module.exports = function (req, res) {
...
+                return postRegistrationHandler(expandedAccount, req, res, defaultJsonResponse.bind(null, req, res));
+              }
+
+              defaultJsonResponse(req, res);
             });
+          });
+        });
+      });
+    },
+    // Handle incoming POST requests from an browser-like clients (something like
```
timothyej: You're handling both `POST` and `GET` here.

lib/controllers/register.js
```diff
@@ -125,159 +125,174 @@ module.exports = function (req, res) {
...
 
-  // Handle incoming POST requests from an API-like clients (something like
-  // Angular / React / REST).
-  if (req.method === 'POST' && accepts === 'json') {
-    applyDefaultAccountFields(config, req);
-    helpers.validateAccount(req.body, config, function (errors) {
-      if (errors) {
-        return res.status(400).json({ error: errors[0].message });
+  helpers.handleAcceptRequest(req, {
+    // Handle incoming POST requests from an API-like clients (something like
```
timothyej: I would also remove the `POST` word here, even though it currently only handles `POST` requests.

lib/controllers/register.js
```diff
@@ -125,159 +125,174 @@ module.exports = function (req, res) {
...
             });
+          });
+        });
+      });
+    },
+    // Handle incoming POST requests from an browser-like clients (something like
+    // chrome / firefox / etc.).
+    'text/html': function () {
+      switch (req.method) {
+        // We should render the registration template.
```
timothyej: This comment is not a bit off, since it is either sending back the SPA file, or it renders the template.

timothyej:
I've verified that this works well :+1: Please see my comments regarding the code comments. I know you didn't write them, but they are no longer correct. Remove them or fix them as suggested.

Also as a side note, this was quite hard to review due to all the indentation changes. And in the future, we should *really* break this up into smaller pieces, it's kind of a mess right now :/

lib/controllers/register.js
robertjd: With #331 , the `handleAcceptRequest` is now serving the SPA root, so this can go away.

robertjd:
This looks good to me, see inline comment about the SPA root now being served differently via #331.

> And in the future, we should really break this up into smaller pieces, it's kind of a mess right now :/

I'm hearing "smaller pieces" a lot, and I'd like to know what your desires and solutions are.  After this milestone, I'd love it if you can prepare some suggestions that we can review.

timothyej:
> I'm hearing "smaller pieces" a lot, and I'd like to know what your desires and solutions are. After this milestone, I'd love it if you can prepare some suggestions that we can review. – @robertjd 

Yaaay, I'd be very happy to :)

lib/controllers/register.js
```diff
@@ -125,173 +125,178 @@ module.exports = function (req, res) {
...
-  // If a GET request is made with the Accept header as json,
-  // respond with the view model for this form.
-  // TODO: When `stormpath.web.produces` is implemented,
-  // this `if` must check that it is set to `application/json`.
-  if (req.method === 'GET' && accepts === 'json') {
-    return helpers.getFormViewModel('register', config, function (err, viewModel) {
-      if (err) {
-        return res.status(400).json({ error: err.userMessage || err.message });
-      }
+  helpers.handleAcceptRequest(req, {
```
timothyej: We should send in `res` here too now, right?

typerandom:
@timothyej Will you review/merge?

timothyej:
I was just waiting for the build to finish, merging now :+1:

PR #333, opened 2016-02-17
Add produces option for /forgot

### LINKED ISSUE #326: Implement produces option for /forgot
Implement the produces option for `/forgot` according to the [framework spec](https://github.com/stormpath/stormpath-framework-spec/blob/master/password-reset.md#request-handling):

>##### GET requests with Accept: text/html:

>If stormpath.web.spa.enabled is false, render a form which allows the user to request a password reset by entering their email address. If the request contains the query ?status=invalid_sptoken, a message should be shown above the form:

>The password reset link you tried to use is no longer valid. Please request a new link from the form below. If stormpath.web.spa.enabled is true, return the SPA view.

>##### POST Requests

>This endpoint accepts a post from the password reset request form, and the only field in this form is the email field. The endpoint should accept the content as application/json or application/x-www-form-urlencoded.

Part solution of #298.

timothyej:
Implements the produces option for `/forgot` according to the framework spec.

#### How to verify

1. Checkout this branch.
4. Start the server.
5. Navigate to `http://localhost:3000/forgot`.
6. Submit your email.
7. Open up the terminal and execute `$ http POST localhost:3000/forgot email=[your name]@stormpath.com --json`.
8. Verify that you get a `200 OK` response.
9. Modify your configuration and change `web.produces` to `['text/html', '']` (hack because extender currently merges arrays... which needs to be fixed, empty string overwrites second array item.).
10. Verify that step 5-6 works, but that step 7 responds with a `404 Not Found` status.
11. Modify the `web.produces` config to `['application/json', '']`.
12. Verify that step 5 results in a `404 Not Found` and that steps 7-8 work as previous.
13. Modify the `web.produces` config again to `['', '']`.
14. Verify that both steps 5 and 7 result in a ´404 Not Found`.

Fixes #326.

### DIFF
--- a/lib/controllers/forgot-password.js
+++ b/lib/controllers/forgot-password.js
@@ -27,7 +27,7 @@ module.exports = function (req, res, next) {
 
   res.locals.status = req.query.status;
 
-  helpers.handleAcceptRequest(req, {
+  helpers.handleAcceptRequest(req, res, {
     // Handle incoming requests from an API-like clients (something like
     // Angular / React / REST).
     'application/json': function () {
--- a/lib/controllers/forgot-password.js
+++ b/lib/controllers/forgot-password.js
@@ -17,9 +17,9 @@ var helpers = require('../helpers');
  *
  * @param {Object} req - The http request.
  * @param {Object} res - The http response.
+ * @param {function} next - The next callback.
  */
-module.exports = function (req, res) {
-  var accepts = req.accepts(['html', 'json']);
+module.exports = function (req, res, next) {
   var application = req.app.get('stormpathApplication');
   var config = req.app.get('stormpathConfig');
   var logger = req.app.get('stormpathLogger');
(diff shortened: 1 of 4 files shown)


lib/controllers/forgot-password.js
```diff
@@ -27,45 +27,58 @@ module.exports = function (req, res) {
...
-    error: function (form) {
-      // https://github.com/caolan/forms/issues/96
-      if (req.query.status === 'invalid_sptoken') {
-        return helpers.render(req, res, view, { form: form, status: req.query.status });
+
+    // Handle incoming requests from an browser-like clients (something like
+    // chrome / firefox / etc.).
+    'text/html': function () {
+      if (req.method === 'GET' && config.web.spaRoot) {
+        return res.sendFile(config.web.spaRoot);
```
robertjd: This can be removed, as #331 is now handling the serving of the SPA root

robertjd:
+1, looks good!  Please see note about serving the SPA root.

lib/controllers/forgot-password.js
timothyej: This is now fixed, thank :+1:

lib/controllers/forgot-password.js
```diff
@@ -27,45 +27,54 @@ module.exports = function (req, res) {
...
-      }
 
-      var formErrors = helpers.collectFormErrors(form);
-      helpers.render(req, res, view, { form: form, formErrors: formErrors });
-    },
-    // If we get here, it means the user is doing a simple GET request, so we
-    // should just render the forgot password template.
-    empty: function (form) {
-      helpers.render(req, res, view, { form: form });
+    // Handle incoming requests from an browser-like clients (something like
```
typerandom: Remove this comment. It doesn't really add any value. Just bloat.

lib/controllers/forgot-password.js
```diff
@@ -27,45 +27,54 @@ module.exports = function (req, res) {
 
   res.locals.status = req.query.status;
 
-  if (req.method === 'POST' && accepts === 'json') {
-    return application.sendPasswordResetEmail(req.body.email, function (err) {
-      if (err) {
-        logger.info('A user tried to reset their password, but supplied an invalid email address: ' + req.body.email + '.');
+  helpers.handleAcceptRequest(req, res, {
+    // Handle incoming requests from an API-like clients (something like
```
typerandom: Remove this comment. It doesn't really add any value. Just bloat.

lib/controllers/forgot-password.js
```diff
@@ -27,45 +27,54 @@ module.exports = function (req, res) {
...
-      helpers.render(req, res, view, { form: form, formErrors: formErrors });
-    },
-    // If we get here, it means the user is doing a simple GET request, so we
-    // should just render the forgot password template.
-    empty: function (form) {
-      helpers.render(req, res, view, { form: form });
+    // Handle incoming requests from an browser-like clients (something like
+    // chrome / firefox / etc.).
+    'text/html': function () {
+      forms.forgotPasswordForm.handle(req, {
```
typerandom: Should really all request methods be handled here? I think that you only need to handle GET, maybe POST.

lib/controllers/forgot-password.js
timothyej: Removed.

lib/controllers/forgot-password.js
timothyej: Removed.

lib/controllers/forgot-password.js
timothyej: You're right, it will now only handle `GET` and `POST` requests.

typerandom:
@timothyej This is failing. Are you looking into it?

robertjd:
I rebased `produce-option-change` against `framework-spec-upgrade`, then rebased this branch against `produce-option-change`.  As such, this branch now represents the upgrades to both of those routes.  The tests are interdependent on each other, so this makes the most since.  I added a commit to update the SPA test fixture to use the new configuration format.

Tests are good, merging.

timothyej:
Thanks a lot @robertjd! :)

PR #335, opened 2016-02-18
Add produces option for /change

### LINKED ISSUE #328: Implement produces option for /change
Implement the produces option for `/change` according to the [framework spec](https://github.com/stormpath/stormpath-framework-spec/blob/master/password-reset.md#request-handling):

>##### GET requests with Accept: text/html:

>If stormpath.web.spa.enabled is false, render a form which allows the user to request a password reset by entering their email address. If the request contains the query ?status=invalid_sptoken, a message should be shown above the form:

>The password reset link you tried to use is no longer valid. Please request a new link from the form below. If stormpath.web.spa.enabled is true, return the SPA view.

>##### POST Requests

>This endpoint accepts a post from the password reset request form, and the only field in this form is the email field. The endpoint should accept the content as application/json or application/x-www-form-urlencoded.

Part solution of #298.

timothyej:
Implements the produces option for `/change` according to the framework spec.

### How to verify

1. Checkout this branch
4. Start the server
5. Verify that you can complete the forgot-password flow (testing especially `/change`)
6. Set `web.produces` to `['text/html', '']`
7. Verify that you still can complete the forgot-password flow
8. Set `web.produces` to `['application/json', '']`
9. Got to <http://localhost:3000/change?sptoken=sdsd> and verify that you get json back
10. Using Postman or another tool, post to <http://localhost:3000/change?sptoken=token> with a real token and `{ "password": "pwd" }`. Also set Accept to `application/json`
11. Verify that you can change your password

Fixes #328.

### DIFF
--- a/lib/controllers/change-password.js
+++ b/lib/controllers/change-password.js
@@ -30,19 +30,16 @@ module.exports = function (req, res, next) {
     return res.redirect(config.web.forgotPassword.uri);
   }
 
-  application.verifyPasswordResetToken(sptoken, function (err, result) {
-    if (err) {
-      logger.info('A user attempted to reset their password with a token, but that token verification failed.');
-    }
+  helpers.handleAcceptRequest(req, res, {
+    'application/json': function () {
+      if (req.method !== 'POST') {
+        return next();
+      }
 
-    helpers.handleAcceptRequest(req, res, {
-      'application/json': function () {
+      application.verifyPasswordResetToken(sptoken, function (err, result) {
         if (err) {
+          logger.info('A user attempted to reset their password with a token, but that token verification failed.');
           return res.status(400).json({ error: err.userMessage || err.message });
-        }
-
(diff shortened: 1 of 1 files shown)


lib/controllers/change-password.js
```diff
@@ -33,66 +33,80 @@ module.exports = function (req, res) {
   application.verifyPasswordResetToken(sptoken, function (err, result) {
```
typerandom: This needs to be inside the `helpers.handleAcceptRequest()`, else you'll handle the request even if `web.produces` is empty.

timothyej: Thanks, I just fixed this :+1:

robertjd:
PR #333 was rebased against this branch (which was also rebased against `framework-spec-upgrade`).  As such, this branch is now included by #333 so I am closing.

PR #337, opened 2016-02-19
Changes to support the produces option

robertjd:
**Description of changes:**

* Determine the correct content type response, based on accept header preferences and produces configuration

* New test suite to assert that the correct content type is being served, given the supplied `stormpath.web.produces` configuration.

* Consolidate SPA serving into a single place.

**Verification**

Run the new tests :) Or checkout the the [express-stormpath-sample-project](https://github.com/stormpath/express-stormpath-sample-project) and modify the `web.produces` configuration with various options, and assert that you get the correct content types when you change the Accept header for your requests.

**Meta/Discuss**

I'm happy with the new test suite, but not so happy with what's going on inside `handleAcceptRequest`.  Dealing with the SPA serving inside of there isn't right, we should break that out somewhere else.  Doing so will allow us to clean up the content type delegation.  This can be done as a refactor in the future.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -222,14 +222,16 @@ React.  For each feature (login, registration) there is a JSON API for the
 feature.  The JSON API is documented for each feature, please see the feature
 list in the sidebar of this documentation.
 
-In some cases you may need to specify the ``spaRoot`` option.  This
+In some cases you may need to specify the ``spa.view`` option.  This
 is the absolute file path to the entry point for your SPA.  That option
 would be defined like this::
 
     app.use(stormpath.init(app, {
-      website: true,
       web: {
-        spaRoot: path.join(__dirname, 'public', 'index.html')
+        spa: {
+          enabled: true,
+          view: path.join(__dirname, 'public', 'index.html')
+        }
       }
     }));
 
--- a/lib/helpers/handle-accept-request.js
+++ b/lib/helpers/handle-accept-request.js
@@ -1,31 +1,75 @@
 'use strict';
 
+var _ = require('lodash');
+
+var SpaResponseHandler = require('./spa-handler');
+
 /**
- * Handles an HTTP Accept request.
+ * Determines which handler should be used to fulfill a response, given the
+ * Accept header of the request and the content types that are allowed by the
+ * `stormpath.web.produces` configuration.  Also handles the serving of the SPA
+ * root page, if needed.
  *
  * @method
  * @private
  *
  * @param {Object} req - HTTP request.
  * @param {Object} res - HTTP response.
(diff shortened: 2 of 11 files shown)


lib/helpers/handle-accept-request.js
```diff
@@ -1,31 +1,74 @@
...
+  var allowedResponseTypes = _.intersection(produces, accepted);
+
+  /*
+    Of the allowed response types, find the first handler that matches. But
+    always override with the SPA handler if SPA is enabled.
+  */
+
+  var handler;
+
+  allowedResponseTypes.forEach(function (contentType) {
```
timothyej: You can use `.some()` or `.find()` here instead and then return `true` to abort iteration.

lib/helpers/handle-accept-request.js
```diff
@@ -1,31 +1,74 @@
...
+    if (handler) {
```
timothyej: If you go with `.some()` or `.find()` you won't need this check here and can remove this `if` statement :)

timothyej:
It works fine with the `express-stormpath-sample-project`, but it doesn't work with the `stormpath-express-react-example`. When trying to login or signup, the response is `404` no matter what I set produces to.

timothyej:
Ignore my previous comment. Everything works just fine, it was just something wrong with my `npm`.

timothyej:
I've verified that everything works as expected. Except for my minor comment, the code looks nice, and the tests looks really nice too :+1: 

I've updated the README with the new `web.spa` options (fe583a047cb835a7c9da6d8a49a516d4b31dbebf).

typerandom:
:+1: Also works for me. I have updated the docs to fix another `spaRoot` reference.

docs/configuration.rst
```diff
@@ -222,14 +222,14 @@ React.  For each feature (login, registration) there is a JSON API for the
...
-In some cases you may need to specify the ``spaRoot`` option.  This
+In some cases you may need to specify the ``spa.view`` option.  This
 is the absolute file path to the entry point for your SPA.  That option
 would be defined like this::
 
     app.use(stormpath.init(app, {
-      website: true,
-      web: {
-        spaRoot: path.join(__dirname, 'public', 'index.html')
+      spa: {
```
timothyej: This should be inside the `web` config.

typerandom:
@timothyej I fixed your comments and cleaned up a few things. Can you do a quick review so that we can merge?

typerandom:
@robertjd Thanks for fixing the issue with this PR.

timothyej:
:+1:

robertjd:
Woot!  Thanks all for the array fixes and doc fixes :)

PR #340, opened 2016-02-19
Upgrade JSON error responses

### LINKED ISSUE #325: Upgrade JSON error responses
Upgrade JSON error responses according to the [framework spec](https://github.com/stormpath/stormpath-framework-spec/blob/master/error-handling.md#error-handling).

>The only properties that should ever be exposed to the end-user are the message and status properties. As such, this error would be sent to the end-user like so:
>```json
>{
>  "status": 400,
>  "message": "Invalid username or password."
>}
>````

#### Discussion

This will be a breaking change since the JSON error field has changed from `error` to `message`. How should we handle this?

I.e. [https://github.com/stormpath/express-stormpath/blob/master/lib/controllers/register.js#L104](https://github.com/stormpath/express-stormpath/blob/master/lib/controllers/register.js#L104):

```json
{
  "error": "Invalid username or password."
}
```

Will according to the new framework spec be:

```json
{
  "status": 400,
  "message": "Invalid username or password."
}
```

Part of #298.

timothyej:
This PR upgrades JSON error responses according to the [framework spec](https://github.com/stormpath/stormpath-framework-spec/blob/master/error-handling.md#error-handling).

It doesn't handle any Node errors that might occur due to e.g. network errors and bugs. These should be caught and sent as status `500 Internal Server Error`. This will have to be addresses in another PR.

Fixes #325.

### Notes

It can be a bit hard to actually verify all errors, so try a few and then put in some extra time into reviewing the code.

### How to verify

1. Checkout this branch
2. Use Postman or another HTTP tool and set the Accept header to `application/json`
3. Call the API and try to provoke different errors (example on `/login` and `/register`)
4. Verify that the errors have the same format as in the spec:

  ```javascript
{
    "status": 400,
    "message": "First Name required."
}
```

### CODE REVIEW

**Comment by timothyej** [comment]
I will look into the failing tests.

**Comment by timothyej** [comment]
The tests are now passing. I had missed that`/oauth/token` errors should look a little different:

> Error responses from the `/oauth/token` endpoint should pass ONLY the `error` and `message` properties.
> – [Error Handling Spec](https://github.com/stormpath/stormpath-framework-spec/blob/master/error-handling.md#error-handling)

File: lib/controllers/get-token.js [hunk 1]
```diff
@@ -21,8 +21,14 @@ module.exports = function (req, res) {
   var logger = req.app.get('stormpathLogger');
 
   function writeErrorResponse(err) {
+    var error = {
```
typerandom: For consistency, please use `helpers.writeJsonError(res, err)` here instead.

typerandom:
There are a few places that you've missed:
* https://github.com/stormpath/express-stormpath/blob/json-errors/lib/middleware/api-authentication-required.js#L32
* https://github.com/stormpath/express-stormpath/blob/json-errors/lib/middleware/api-authentication-required.js#L38
* https://github.com/stormpath/express-stormpath/blob/json-errors/lib/middleware/api-authentication-required.js#L44
* https://github.com/stormpath/express-stormpath/blob/json-errors/lib/middleware/authentication-required.js#L34
* https://github.com/stormpath/express-stormpath/blob/json-errors/lib/controllers/get-token.js#L31
* https://github.com/stormpath/express-stormpath/blob/json-errors/lib/controllers/register.js#L260
* https://github.com/stormpath/express-stormpath/blob/json-errors/lib/helpers/login-with-oauth-provider.js#L29
* https://github.com/stormpath/express-stormpath/blob/json-errors/lib/helpers/xsrf-validator.js#L18

Besides this, the PR is looking good! :)

timothyej:
Thanks @typerandom, that's a few ;) However, the `get-token.js` should not be changed, see my previous comment about that.

lib/controllers/get-token.js
timothyej: Please see my comment about that: https://github.com/stormpath/express-stormpath/pull/340#issuecomment-187131588

timothyej:
> * https://github.com/stormpath/express-stormpath/blob/json-errors/lib/middleware/api-authentication-required.js#L32

Fixed :+1: 

> * https://github.com/stormpath/express-stormpath/blob/json-errors/lib/middleware/api-authentication-required.js#L38

Fixed :+1: 

> * https://github.com/stormpath/express-stormpath/blob/json-errors/lib/middleware/api-authentication-required.js#L44

Fixed :+1: 

> * https://github.com/stormpath/express-stormpath/blob/json-errors/lib/middleware/authentication-required.js#L34

Fixed :+1: 

> * https://github.com/stormpath/express-stormpath/blob/json-errors/lib/controllers/get-token.js#L31

Not missed, this is according to the spec (OAuth errors are different).

> * https://github.com/stormpath/express-stormpath/blob/json-errors/lib/controllers/register.js#L260

I don't know what to do with this one as the Accept type if `text/html` but the error response is still JSON. And also, it's an array of errors. I don't really want to touch this one as I don't know what I can mess up (I'm the one who added the comment btw).

> * https://github.com/stormpath/express-stormpath/blob/json-errors/lib/helpers/login-with-oauth-provider.js#L29

Fixed :+1: 

> * https://github.com/stormpath/express-stormpath/blob/json-errors/lib/helpers/xsrf-validator.js#L18

Fixed :+1: (But it seems like this file isn't used at all?)

lib/controllers/register.js
```diff
@@ -270,6 +256,7 @@ module.exports = function (req, res, next) {
                   password: req.body.password || uuid()
                 }, function (err, passwordGrantAuthenticationResult) {
                   if (err) {
+                    // NOTE: Why is there a JSON error when the Accept header is text/html?
```
typerandom: This should probably be:

```javascript
logger.info(err);
return helpers.render(req, res, view, {
	errors: [new Error(err.userMessage)],
	form: helpers.sanitizeFormData(req.body, config),
	formModel: formModel
});
```

typerandom:
Besides my latest comment! This looks good now! :)

robertjd:
Looks good!  I addressed the error that should be rendered in the from, as @typerandom mentioned.  I also added a few tests to assert that we are rending the email verification errors correctly.  I rebased against current framework-spec-upgrade branch.

timothyej:
Awesome, thanks @robertjd!

PR #345, opened 2016-02-22
Use local token validation by default

### LINKED ISSUE #315: Use local token validation by default
In our 3.0 release, we will start using using [local token validation](http://docs.stormpath.com/nodejs/express/latest/authentication.html#token-validation-strategy) by default.  This should be a simple change to the default [web-config.yaml](https://github.com/stormpath/stormpath-framework-spec/blob/master/web-config.yaml)

Part of #298.

timothyej:
This PR changes the default configuration to use [local token validation](http://docs.stormpath.com/nodejs/express/latest/authentication.html#token-validation-strategy). It also updates the documentation accordingly.

### How to verify

You can always test that the token validation is actually local, but since there are no code changes is this PR, a review of the configuration and documentation changes would be sufficient.

Fixes #315.

### DIFF
--- a/docs/authentication.rst
+++ b/docs/authentication.rst
@@ -143,9 +143,11 @@ Opt-in to ``stormpath`` validation with this configuration:
 
   When using local validation, your server will not be aware of token revocation
   or any changes to the associated Stormpath account.  **This is a security
-  risk.**
+  tradeoff that optimizes for performance.**  If you prefer extra security, use
+  the ``stormpath`` validation option.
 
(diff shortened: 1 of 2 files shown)


typerandom:
Looks good to me! :) +1

robertjd:
Looks good, thanks!  I added a small clarification about the performance vs. security tradeoffs.  When I revisit all the docs for the 3.0 release, I think I will add some information to the quickstart section and reference this area of the docs.

PR #346, opened 2016-02-22
Don't attach the middleware in stormpath.init()

### LINKED ISSUE #285: Don't apply `getUser` middleware to passed app
By default, we will attach the `getUser` middleware to all requests on the application or router instance that is passed into `stormpath.init()`, as seen here:

https://github.com/stormpath/express-stormpath/blob/master/lib/stormpath.js#L219

As such, we force all requests through our authentication layer and this adds network overhead when you're logged in.  If you accidentally apply our module to your "public" folder (static assets, etc), a folder that doesn't require authentication, you app becomes unexpectedly slow.  Myself and other develoeprs have run into this many times.

Should we be doing this?  

My opinion is that `stormpath.init()` not apply authentication middleware.  We provide explicit middleware functions, such as `stormpath.loginRequired`.  As such we should allow the developer to tell us what routes that they want protected.

timothyej:
**WIP - Don't merge, see Discussion!**

Fixes #285.

**Before this fix**, `stormpath.init()` would both attach the middleware and then return it:

```javascript
   app.use(stormpathMiddleware);
 
   return stormpathMiddleware;
 };
```

**With this fix**, `stormpath.init()` returns a `nopMiddleware` which is used for backwards compatibility when `stormpath.init()` is wrapped with an `app.use()`. Then the `stormpathMiddleware` is only attached for the Stormpath specific routes:

```javascript
function addGetRoute(path, controller) {
  router.get(path, bodyParser.forceDefaultBody(), stormpathMiddleware, controller);
}

function addPostRoute(path, controller, options) {
  router.post(path, bodyParser.formOrJson(options), stormpathMiddleware, controller);
}
```

### Discussion

Problem with this is that the menu/navbar stopped working in the example app. The reason is that it need to know whether the user is logged in or not to show the login or the logout button.

```jade
ul.nav.navbar-nav.navbar-right
  if user
    li(class='#{nav.Logout}')
      a(href="/logout") Logout
  else
    li(class='#{nav.Login}')
      a(href='/login') Login
```

One solution would be to manually attaching a new stormpath middleware that would get the user. **An easier solution would be to still return the `stormpathMiddleware` and use `app.use()` on `stormpath.init()`** (that's what the example app does).

### DIFF
--- a/lib/stormpath.js
+++ b/lib/stormpath.js
@@ -90,6 +90,12 @@ module.exports.init = function (app, opts) {
     helpers.getUser(req, res, next);
   }
 
+  // This middleware is used for backwards compatibility
+  // when stormpath.init() is wrapped with an app.use().
+  function nopMiddleware(req, res, next) {
+    next();
+  }
+
   // Build routes.
   client.on('ready', function () {
     var config = app.get('stormpathConfig');
@@ -118,11 +124,11 @@ module.exports.init = function (app, opts) {
     }
 
     function addGetRoute(path, controller) {
-      router.get(path, bodyParser.forceDefaultBody(), controller);
+      router.get(path, bodyParser.forceDefaultBody(), stormpathMiddleware, controller);
     }
 
     function addPostRoute(path, controller, options) {
-      router.post(path, bodyParser.formOrJson(options), controller);
+      router.post(path, bodyParser.formOrJson(options), stormpathMiddleware, controller);
     }
 
     router.use(localsMiddleware);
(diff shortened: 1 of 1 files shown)


timothyej:
FYI: I think I will revert some of this and just do as the issue title suggests; "Don't apply `getUser` middleware to passed app". That means I would still have to return the `stormpathMiddleware` and I could remove the `nopMiddleware`.

@typerandom @robertjd Let me know what you think.

robertjd:
@timothyej I want to think about this more and look at this again in the morning.

timothyej:
@robertjd Sure, take your time :)

robertjd:
Closing in favor of #358, thanks for getting this one started.

PR #349, opened 2016-02-24
Add support for stormpath.web.social configuration options

### LINKED ISSUE #316: Add support for stormpath.web.social configuration

timothyej:
This PR implements the `stormpath.web.social` configuration options.

* Adds `web.social` to the default configuration
* Renames `config.socialProviders` to `config.web.social`
* Renames `.callbackUri` to `.uri` for those configurations
* The `/spa-config` route (soon to be deprecated) should only return enabled providers
* Solved a circular dependency

This PR is dependent on the `master` of [stormpath-node-config](https://github.com/stormpath/stormpath-node-config).

Fixes #316.

### How to verify

1. Make sure that the config parameters and its default values looks correct
2. Checkout `master` of `stormpath-node-config` and link it
3. Checkout this branch and link `stormpath-node-config`
4. Link `express-stormpath` to your example project
5. Run your example project
6. Try to login with different social providers


timothyej:
Tests fails because it depends on `master` of `stormpath-node-config`. We should release a new version of that and then upgrade this PR to use it.

robertjd:
@timothyej The templates needed to be updated to pass the scope value directly.  I made that change and I'm merging now.

timothyej:
Awesome, thanks :)

PR #360, opened 2016-02-26
Fix UNKNOWN override when name field optional

### LINKED ISSUE #329: Making givenName and surname optional should not set fields to UNKNOWN
If I set givenName/surname to not default, express sets it to UNKNOWN regardless of the input. This should not happen.

For instance, if I set: 

```
web: {
register: {
  enabled: true,   // Explicit enable, if not using { website: true }
  nextUri: '/',    // Where to send the user to, if auto login is enabled
  fields: {
    givenName: {
      required: false
    },
```

And register with a Given Name set, `express-stormpath` turns my givenName into "UNKNOWN"

https://github.com/stormpath/express-stormpath/blob/master/lib/controllers/register.js#L84

robertjd:
If the user supplies the value, but the field is optional, we should not override with UNKNOWN

Closes #329

### DIFF
--- a/lib/controllers/register.js
+++ b/lib/controllers/register.js
@@ -84,11 +84,11 @@ function defaultJsonResponse(req, res) {
 function applyDefaultAccountFields(stormpathConfig, req) {
   var registerFields = stormpathConfig.web.register.form.fields;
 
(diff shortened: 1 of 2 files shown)


edjiang:
w00t w00t

PR #366, opened 2016-03-02
Fix cookies for social login

robertjd:
During the yaml upgrades, we forgot to default the cookie path to / if not specified.  This is required because if we omit the path, the cookies will be set on the /callback URI when social login happens, and those cookies are then trapped on that URL and can’t be used on the rest of the site.

I’ve added a test case in the login controller test, to assert that we set the cookies with the default path of /

I’ve also refactored that test file to create two express apps fixtures in the begin block, which is must faster than creating one at every block.  This is a test change that I’ve been implementing in most of my PRs.

I’ve added some clarifying comments to the yaml file, about cookies. Ideally we would have some actual IT tests to assert social callbacks, but this new test will work for now because we have a common method for creating our cookies.

Once merged I’ll release this at 3.0.1

### DIFF
--- a/lib/config.yml
+++ b/lib/config.yml
@@ -32,7 +32,7 @@ web:
     secure: null
 
     # Controls the path flag of the cookie.  Inherits from basePath, but can be
-    # over-ridden here.
+    # overridden here.
     path: null
 
     # Controls the domain flag on the cookie, will not be set unless specified.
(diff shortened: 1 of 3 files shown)


timothyej:
Nice work! I just fixed a little typo, merging now :+1:

PR #368, opened 2016-03-02
pre-warm view model cache for faster SPA applications

robertjd:
Quick change to get these cached warmed, so that you're not waiting on them when you load your Angular application for the first time.  Not a huge fan of the anon callbacks, can refactor later if desireed.


timothyej:
Works fine :+1:

PR #370, opened 2016-03-03
Validate the hosted domain parameter (hd) when used on the initial oa…

bosticka:
Validate the hosted domain parameter (hd) when used on the initial oauth uri.

Just specifying hd=yourdomain.com is not enough to ensure only your domain
authenticates.

Google returns the authenticated hosted domain as a query parameter on the
callbackURI.  This value must match the hd value intially supplied or its a
waste of time and any domain may authenticate.

See also: https://developers.google.com/identity/protocols/OpenIDConnect#hd-param

To test this, use something like:
```
app.use(stormpath.init(app, {
    ...
    socialProviders: {
        google: {
            hd: "yourdomain.com",
            callbackUri: "/your/callback"
        }
    }
}));
```

in your main express app.js file.

### DIFF
--- a/lib/controllers/google-login.js
+++ b/lib/controllers/google-login.js
@@ -30,6 +30,14 @@ module.exports = function (req, res) {
   var logger = req.app.get('stormpathLogger');
   var loginHandler = config.postLoginHandler;
   var registrationHandler = config.postRegistrationHandler;
+  var socialProviders = config.socialProviders;
+
+  if (socialProviders.google.hd) {
+    if (!req.query.hd || req.query.hd != socialProviders.google.hd) {
+      logger.info('A user attempted to log in via Google OAuth using an invalid hosted domain.');
+      return res.status(400).json({ message: 'invalid hosted domain parameter.' });
+    }
+  }
 
   if (!req.query.code) {
     logger.info('A user attempted to log in via Google OAuth without specifying an OAuth token.');


bosticka:
The original commit was for the 2.5  branch of express-stormpath.  The newest commit updates to the 3.0 release since there are breaking changes involved.

robertjd:
Hello @bosticka , thanks for the PR!  JFYI, the tests will always fail on this PR because it's coming from outside of our repo.  If we decide to move forward with the PR, we have to copy the fork into our repo and issue a second PR.

All that said, a question about the implementation: The google documentation says `be sure to validate that the returned ID token has a matching hd claim`.  Are we supposed to look into the token?  In this PR I see us looking at query params and not a token.

bosticka:
Hi Robert,

I think the wording is just a little ambiguous.  The callback uri response
contains the token (refresh and access as response headers) of the
authorized user but also other things about the initial request as query
params.  Some of the fields, such as state, are round trip fields, google
just sends back what you sent.

State, for example, is meant to be a place where you can sign your initial
request to preclude tampering (man in the middle) on the response.

My initial thought was the hd parameter on the request would prevent
authentication.   It does not, and google apparently does not do much with
it (I think they did something with it on their older deprecated
protocols?).  However, they send their hd query param on the response with
the domain of the authenticated user.

Since this request comes straight from google, it think its totally ok to
just match query params on the response vs request.

This use case is for google apps hosted domains (what I have) versus just
generic gmail or other gmail hosted domain logins.

I've tested this in my own environment and its working well on my forked
repo.

Hope this answers your question.

Aaron

bosticka:
Any chance to get this merged?  Anyone using this library with a google hosted domain is wide open to authentication from *any* gmail account without this check...  It's a huge security hole!

robertjd:
Hi @bosticka , thanks for the bump.  I'll pull this up for us to look at.  I'm concerned that the HD parameter could be spoofed by a malicious agent (because it's a simple GET param), so I want us to look at the ID token assertion (as mentioned above) to make sure that we're making this as secure as possible.

PR #371, opened 2016-03-04
Fix: registration fields omitted if not in fieldOrder

robertjd:
We want the developer to be able to specify a new custom field, without having to explicitly enumerate a new fieldOrder array.  Like this:

```
app.use(stormpath.init(app, {
  web: {
    register: {
      form: {
        fields: {
          favoriteColor: {
            enabled: true,
            label: 'Favorite Color',
            name: 'favoriteColor',
            placeholder: 'E.g. Red, Blue',
            required: true,
            type: 'text'
          }
        }
      }
    }
  }
}));
```

This was working for our HTML views but not our JSON views.  This PR fixes that problem by patching the new view model builder, and deprecating the old HTML one.  Now we use the same for both :)

P.S. Thanks to @timothy for the new view model builder, I really like this one.

Note that I had to change the way that we resolve the TTL for the cache.  The Node SDK does not support `config.cacheManager.defaultTtl` yet.  As such, I’ve patched this library to look for `config.cacheOptions.ttl`, as that is how the developer would be passing it in, given the way we have this documented right now.

I also had to modify the registration test to bypass the cache, because we’re using an in-process memory cache that isn’t cleared across test runs.  This is the downside of doing caching at a layer above the Node SDK, but this is fine until we implement collection caching in the Node SDK.

The changes in the registration controller look epic but it’s really just a nesting change: everything is now moved to the right by one indentation, so that it’s all wrapped in a call to get the registration view model.

Woot!


robertjd:
Looks good!  Verified with the `feature-profile-support` branch in the sample project, and the account-save fix branch in the Node SDK.

robertjd:
Lol sh* wrong tab. +1 for self review!

typerandom:
@robertjd Haha!

typerandom:
I have reviewed this and so far it's looking good. I just have one issue and that is that if I set the following config:

```json
{
    "web": {
        "register": {
            "form": {
                "fields": {
                    "existingPassword": {
                        "enabled": true,
                        "label": "Existing Password",
                        "placeholder": "Existing Password",
                        "required": true,
                        "type": "password"
                    }
                },
                "fieldOrder": ["existingPassword"]
            }
        }
    }
}
```

My custom field ends up at the bottom, whereas I'd rather have it respect the field that I provided first and then just put the rest of them at the bottom.

If I explicitly specify all of the enabled fields, it all works fine. But for this little case, I'd say that its behaviour isn't really as expected.

@robertjd What do you think?

typerandom:
Added a tweak for missing fields in fieldOrder. So the PR now looks good so I'm going to merge that in.

PR #381, opened 2016-03-10
Updates to the Setup and Configuration documentation

robertjd:
Previously, the steps required to get started with this module were split between the “Setup” page and the “Configuration” page.

This PR fixes that strangeness.  The “Setup” page is now quickstart-like, and the “Configuration” page has more in-depth information about how to configure the library, and what the defaults are.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -76,7 +76,7 @@ Default Features
 ----------------
 
 When you add Stormpath to your application using ``app.use(stormpath.init(app))``,
-our module will automatically add the following routes to your applicationx:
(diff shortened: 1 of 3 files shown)


typerandom:
I just reviewed this and made a few edits. Besides my edits, the changes looks good and can be merged! Good job! :)

robertjd:
Thanks for the review and edits!  FYI, I reverted the changes to the sample application initialization because (a) I want to demonstrate the stormpath.ready event and (b) the callback for app.listen() doesn't actually receive an error, to my knowledge you have to use `server.on('error')` to capture those events.

PR #387, opened 2016-03-14
Corrections to produces: JSON default, send 406 if no match

### LINKED ISSUE #384: loginRequired middleware doesn't respect produces config
If I set:

```
app.use(stormpath.init(app, {
  web: {
    produces: ['application/json']
  }
}));
```

And I send a HTTP request to a route with `stormpath.loginRequired` without an `Accept` header, Stormpath redirects me to `/login?next=route`

Expected behavior: Stormpath returns `401 Unauthorized`

HTTP logs:

```
GET /notes

HTTP/1.1 302 Found
X-Powered-By: Express
Location: /login?next=%2Fnotes
Vary: Accept
Content-Type: text/plain; charset=utf-8
Content-Length: 44
Date: Fri, 11 Mar 2016 22:54:32 GMT

GET /notes
Accept: application/json

HTTP/1.1 401 Unauthorized
X-Powered-By: Express
Date: Fri, 11 Mar 2016 22:54:46 GMT
```

robertjd:
Fixes #384 

We've finalized the spec for our "produces" option and how we handle content-type negotiation.  You can see the details [here](https://github.com/stormpath/stormpath-framework-spec/blob/master/requests.md#accept-header).

This commit makes two change to bring this library up to date:
* The default response type is JSON if the client does not specify anything.  As such, some of our tests had to be more specific and request text/html when needed.
* We now return 406 if the client asks for an unsupported request type.

The first item is technically a breaking change, for anyone that was relying on a default HTML response when not specifying an Accept preference.  But I don't want to roll up to 4.x for this, so we'll just call it out in the changelog.  I'll be really surprised if anyone is relying on this behavior.

### DIFF
--- a/lib/helpers/handle-accept-request.js
+++ b/lib/helpers/handle-accept-request.js
@@ -44,7 +44,10 @@ function handleAcceptRequest(req, res, handlers, fallbackHandler) {
   // always override with the SPA handler if SPA is enabled.
   var handler;
 
-  allowedResponseTypes.some(function (contentType) {
+  accepted.some(function (contentType) {
+    if (allowedResponseTypes.indexOf(contentType) === -1) {
+      return false;
+    }
     if (config.web.spa.enabled && contentType === 'text/html') {
       handler = spaResponseHandler(config);
       return true;
--- a/lib/helpers/handle-accept-request.js
+++ b/lib/helpers/handle-accept-request.js
@@ -30,11 +30,15 @@ function handleAcceptRequest(req, res, handlers, fallbackHandler) {
   // Our default response is HTML, if the client does not specify something more
   // specific.  As such, map the wildcard type to html.
   accepted = accepted.map(function (contentType) {
-    return contentType === '*/*' ? 'text/html' : contentType;
+    return contentType === '*/*' ? produces[0] : contentType;
   });
 
   // Of the accepted types, find the ones that are allowed by the configuration.
   var allowedResponseTypes = _.intersection(produces, accepted);
+
+  if (!allowedResponseTypes.length) {
+    return res.status(406).end();
+  }
 
   // Of the allowed response types, find the first handler that matches. But
   // always override with the SPA handler if SPA is enabled.
--- a/test/controllers/test-email-verification.js
+++ b/test/controllers/test-email-verification.js
@@ -11,7 +11,8 @@ var SpaRootFixture = require('../fixtures/spa-root-fixture');
 function requestVerifyPage(app, sptoken) {
   var config = app.get('stormpathConfig');
   return request(app)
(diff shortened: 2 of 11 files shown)


typerandom:
Hey, I have some issues with this behaviour...

Instead of:

```javascript
accepted = accepted.map(function (contentType) {
  return contentType === '*/*' ? produces[0] : contentType;
});

// Of the accepted types, find the ones that are allowed by the configuration.
var allowedResponseTypes = _.intersection(produces, accepted);

if (!allowedResponseTypes.length) {
  return res.status(406).end();
}
```

I think it should be:

```javascript
accepted = accepted.map(function (contentType) {
  return contentType === '*/*' ? false : contentType;
});

// Of the accepted types, find the ones that are allowed by the configuration.
var allowedResponseTypes = _.intersection(produces, accepted);

if (!allowedResponseTypes.length) {
  return fallbackHandler();
}
```

The reason, at least from how I have interpreted it. Is that the `produces` option states what we **should** produce. I.e. if it's not listed among what we should produce, then we shouldn't handle it. We should just `next()` it (i.e. in this case call `fallbackHandler()`).

I.e. if you have a look at the React example: https://github.com/stormpath/stormpath-express-react-example/blob/master/server.js#L28

You'll see that the way that we handle SPA is by simply intercepting all requests not handled by the Stormpath middleware. So if it is suddenly handled, then that kind of setup will be broken.

timothyej:
I've started to review this and it breaks the behaviour of the sample-project.

The browser sends the `Accept` header as `text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8` and the config is set to:

```yaml
# By default the Stormpath integration should respond to JSON and HTML
# requests.  If either is removed from configuration, the integration should
# not try to handle the response for the given content type.
produces:
  - application/json
  - text/html
```

My guess is that since it finds `*/*` at the end and replaces it with `application/json`, it will match that when doing the intersection.

Is this correct? I would assume that since the browser lists `text/html` first, that would be the preferred content type?

robertjd:
Oh dang, it did break the sample project (I swore I had tested it..).  Yes, that does mean that the Accept string is being parsed incorrectly.  P.S. Hipchat went down, can you guys join me on slack?

typerandom:
First saw this now. I'm on Hipchat. Did you guys continue talking on slack?

robertjd:
I agree with @typerandom that this configuration isn't ideal, we put together this conversation here:

https://github.com/stormpath/stormpath-framework-spec/issues/62

If we can't resolve this issue soon, we should release 3.1.0 without this PR

robertjd:
I'm closing this PR, as we've decided that this is not the approach that we want.

PR #388, opened 2016-03-15
Fixing JSON response for verifying password reset tokens

### LINKED ISSUE #385: sptoken in password reset link not validated
I'm using express-stormpath 3.0.1 and angular sdk 1.0.0. On loading the password reset page, the angular sdk makes a GET request of the form `http://localhost:9000/change?sptoken=...`. This request always returns the index.html page regardless of the value of sptoken which the angular sdk interprets as a successful response and then shows the password reset form.

The password change submission does not work unless the sptoken value is valid.

robertjd:
In the failure case, we need to render a JSON error message (not redirect to the login page).  Fixes #385 

To verify:

* Use this branch with the [Stormpath Angular + Express Fullstack Sample Project](https://github.com/stormpath/express-stormpath-angular-sample-project)

* Visit http://localhost:3000/change?sptoken=x

With the fix, you should see the error message: "This password reset link is not valid. You may request another link by clicking here." (Expected behaviour)

Without the fix, you should see the change password (problem behaviour).

I've added a new test for HTML vs JSON on this endpoint, and refeactored this test file to use a single default express application (where possible)..

### DIFF
--- a/lib/controllers/change-password.js
+++ b/lib/controllers/change-password.js
@@ -32,14 +32,14 @@ module.exports = function (req, res, next) {
 
   helpers.handleAcceptRequest(req, res, {
     'application/json': function () {
+      if (req.method !== 'GET' && req.method !== 'POST') {
+        return next();
+      }
+
       application.verifyPasswordResetToken(sptoken, function (err, result) {
         if (err) {
           logger.info('A user attempted to reset their password with a token, but that token verification failed.');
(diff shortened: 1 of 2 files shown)


typerandom:
This works as advertised! :) Though I cleaned it up a little bit and changed so that the method only accepts GET and POST requests.

robertjd:
Thanks @typerandom !

PR #393, opened 2016-03-21
Logger fixes

### LINKED ISSUE #383: custom logger is ignored
I'm passing a "logger" parameter (a Winston logger) to stormpath.init, but it looks like it's being ignored and the log keeps going to the console using the fallback defined in lib/stormpath.js line 33.

I think the problem resides in the fact that "stormpathLogger" is read via an "app.get" at line 30, but there was never an "app.set" call before.

For the moment I worked around the issue by explicitly calling "app.set('stormpathLogger", ...)" before calling "stormpath.init", but this looks like a bug to be fixed, if I'm not wrong.

robertjd:
Fixes #383 

Custom loggers were not being used, and only one log level string can be used.

Cleaning up stormpath.init() tests while I’m at it.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -248,13 +248,13 @@ Logging
 By default, this library will create a `Winston`_ logger and use this for
 logging error messages to standard output.
 
-While actively developing your application, you may want to include the ``info``
-level for debugging purposes:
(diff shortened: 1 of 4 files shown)


timothyej:
I've verified that this works and looks good :+1:

PR #403, opened 2016-03-28
more change-password fixes

robertjd:
Another fix for #385.  In #388 we fixed the invalid token case but broke the valid token case in ef8822a13ffb7965b2e308cc37bc3984db766967.

I've added more tests to ensure that we don't break this API in the future, we also need to look at the email verification API to ensure that it is working correctly for JSON and that it has tests.


timothyej:
I've verified that this behaves as expected for both invalid and valid tokens. And I've tried this on both the Angular and Express project. Tests looks good too, nice work :beers:

PR #404, opened 2016-03-29
Pass complete path when delegating to view engine

golmansax:
Fixes bug for when not using jade and template has an extra `.` in its name.  For example, if view is `foo.bar.ejs`, then:
* it calls `res.render('foo.bar')`
* express then strips the `extname` and thinks the template is of type `bar`

I think it's best just to pass the direct value into express, and let it handle everything.  Thoughts?

### DIFF
--- a/lib/helpers/render.js
+++ b/lib/helpers/render.js
@@ -57,7 +57,7 @@ module.exports = function (req, res, view, options) {
     res.send(renderJade(view, options));
   } else if (extension) {
     // Delegate to the view engine.
-    res.render(filename, options);
+    res.render(view, options);
   } else {
     throw new Error('Unexpected view option: "' + view + '".  Please see documentation for express-stormpath');
   }


robertjd:
Hi @golmansax , thanks for this!  We're busy with a few other things this week, but we'll look into this asap!

robertjd:
Hi @golmansax , can you confirm that this change has solved the problem for you?  Even with this change, I get this error from the rendering layer:

> Error: Cannot find module 'bar'

I get the same error even if I try to user render directly (without express-stormpath), like so:

```
'use strict';

var express = require('express');

var app = express();

app.set('view engine', 'ejs');
app.set('views', './views');

app.use('/', function (req, res) {
  res.render('foo.bar');
});

/**
 * Start the web server.
 */
var port = process.env.PORT || 3000;
app.listen(port, function () {
  console.log('Server listening on http://localhost:' + port);
});
```

There is a file named `foo.bar.ejs` in `views`, but I still get the same error.  This makes me think the issue is with the render function in Express?

golmansax:
@robertjd Thanks for getting back!  Yeah, I forgot to mention that I've been calling render with the filename and extension, so with just express it would be `res.render('foo.bar.ejs')`.

However, when using express-stormpath, it calls `res.render('foo.bar')` because it is only uses the filename.  I'm thinking that by passing through the original view, it can maintain any functionality from the original express render method.

typerandom:
I am not able to reproduce this. When I have a view called `foo.bar.ejs` and use the code provided by Robert, then it works perfectly for me.

@golmansax Could you provide code so that I'm able to reproduce this issue?

golmansax:
@typerandom Sure, here is a repo with @robertjd's example https://github.com/golmansax/express-stormpath-demo/blob/master/index.js#L11

typerandom:
@golmansax 

From what I understood of the issue, the problem was that it stripped the extension. So by calling `foo.bar.ejs` it didn't find the template because it looked for `foo.bar`? Or do you want it to automatically look for the template engine (in this case `ejs`) if specified? I.e. if you provide `foo.bar` it will look for `foo.bar.ejs`?

golmansax:
@typerandom Yep, my pull is making it so that it doesn't send the stripped version of the file.  In my app code, I am doing:

```js
stormpath.init(app, {
  web: {
    login: {
      view: 'foo.bar.ejs'
    }
  }
})
```

Which breaks, because it calls `res.render('foo.bar')` in the stormpath code.  https://github.com/stormpath/express-stormpath/pull/404/files#diff-039ba3ca968eb7e233e740a7e70a472dL60

golmansax:
@robertjd thanks!

PR #417, opened 2016-04-16
autoLogin support for client-side apps

peebles:
When a client-side app like react-stormpath does a POST to /change, the only way it can then do an auto-login is to know the account username and password and execute a /login, then a redirect.  The client knows password (because of the change password form), but does not know the username, because the user got to the change password form via an email link.  But if the POST to /change returns the account (which is has), then client can get at the username.

This change was made specifically to support autoLogin capability in react-stormpath.

### DIFF
--- a/lib/controllers/change-password.js
+++ b/lib/controllers/change-password.js
@@ -55,7 +55,7 @@ module.exports = function (req, res, next) {
             return helpers.writeJsonError(res, err);
           }
 
-          res.end();
+          res.json(result);
         });
       });
     },


edjiang:
Hey @peebles, thanks for the pull request! I'm not the maintainer of `express-stormpath`, but I've been working with the team to define behavior (like autologin) on a variety of our integrations.

I looked into how we defined this behavior and it looks like the direction we're going to be moving to across frameworks for autologin is that the `/change` endpoint will set the cookies and also return the `account` object, so you will not have to make a separate call to `/login`. 

See https://github.com/stormpath/stormpath-framework-spec/blob/master/password-reset.md#post-handling

Not sure how the Node team is going to approach this, but I would suggest following that route as well to maintain forward compatibility =]

peebles:
This look perfect.
a

peebles:
This is good to /change.  The other autoLogin I would like to see 
involves /verify.  Here's the use case:

An admin or a script creates users in strormpath ... users cannot self 
register (in this use case).  The admin or script creates a password but 
does not communicate it to the user.  The user receives the verification 
email.  They click on the link.  /verify should verify the account, but 
then should return the account information. The UI presents the account 
information (givenName, surname, email) in a form along with an empty 
password/confirmPassword field.  The user than has a chance to alter the 
account info as input by an admin, and can create a new password.  This 
gets POSTed to /me (?) or some other endpoint, which changes the account 
and optionally does an autoLogin.

This use case is actually arguably more important than the autoLogin 
with /change.

a

edjiang:
Gotcha, so an autologin for `/verify`. I'll bring it up to the team. Thanks!

edjiang:
As an update @peebles, this is now in the plans for the framework integrations like Express, but I'm not quite sure when it'll be actively worked on or out. 

In a future update, enabling `web.register.autoLogin` will also enable it for the `/verify` endpoint, instead of doing nothing. 

You can see the exact changes proposed here: https://github.com/stormpath/stormpath-framework-spec/pull/103

I'm going to close this PR, but thanks for bringing this to our attention!

PR #429, opened 2016-05-09
Fix groupsRequired middleware so that it handles Accept header

### LINKED ISSUE #408: GroupsRequired does not work with JSON
For:

```
app.get('/test', stormpath.groupsRequired(['admins']), function(req, res) {
```

```
GET /test
Accept: application/json

HTTP/1.1 302 Found
X-Powered-By: Express
Location: /login?next=%2Ftest
Vary: Accept
Content-Length: 0
Date: Fri, 01 Apr 2016 19:20:37 GMT
Connection: close
```

I expect to get `401 Unauthorized` instead. 

If I authenticate with a valid access token but without the correct group: 

```
GET /test
Accept: application/json
Authorization: Bearer VALIDTOKEN

HTTP/1.1 200 OK
X-Powered-By: Express
Content-Type: text/html; charset=utf-8
Content-Length: 12181
ETag: W/"2f95-RuQEqVCsRcmcGl7IB/xhmQ"
Date: Fri, 01 Apr 2016 19:23:21 GMT
Connection: close
```
screen shot 2016-04-01 at 12 25 47 pm

I expect to get `403 Forbidden` instead. 

With a user in the correct group and everything, things work as expected.

typerandom:
Fixes the `groupsRequired` middleware so that it responds correctly to the Accept header.

##### How to verify

1. Create a new application using [this gist](https://gist.githubusercontent.com/typerandom/7a85bdaf7792e067c230faea6a56d301/raw/35944ec0d05cd26221f0d475d36fcd83e195fb69/stormpath-express-groups-required-example.js).
2. Without being authenticated, send a `GET` request to `/admin` with a `Accept: text/html` header. This should respond with a location header, redirecting you to the login.
3. Without being authenticated, send a `GET` request to `/admin` with a `Accept: application/json` header. This should respond with a JSON response body and a `401 Unauthorized` status code.
4. Being authenticated, send a `GET` request to `/admin` with a `Accept: text/html` header. This should respond with a location header, redirecting you to the login.
5. Being authenticated, send a `GET` request to `/admin` with a `Accept: application/json` header. This should respond with a JSON response body and a `403 Forbidden` status code.
6. Add the user you are authenticated with in a group called `admin`.
4. Being authenticated in the `admin` group, send a `GET` request to `/admin` with a `Accept: text/html` header. This should respond with a `200 OK` and the JSON response `{message: 'You are an administrator!'}`.
5. Being authenticated in the `admin` group, send a `GET` request to `/admin` with a `Accept: application/json` header. This should respond with a `200 OK` and the JSON response `{message: 'You are an administrator!'}`.

Fixes #408.

### DIFF
--- a/lib/middleware/groups-required.js
+++ b/lib/middleware/groups-required.js
@@ -9,7 +9,7 @@ var helpers = require('../helpers');
  * shown an unauthorized page letting them know they do not have the required
  * permissions.
  *
- * @param {String[]} assertGroups - List of groups to assert membership for. Groups
+ * @param {String[]} groups - List of groups to assert membership for. Groups
  *   must be specified by group name.
  * @param {Boolean} [all=true] - Whether membership is required in one group or all.
  *
@@ -17,11 +17,12 @@ var helpers = require('../helpers');
  *   group membership, and only allows the user to continue if the assertions
  *   are true.
  */
-module.exports = function (assertGroups, all) {
+module.exports = function (groups, all) {
   all = all === false ? false : true;
 
   return function (req, res, next) {
     var config = req.app.get('stormpathConfig');
+    var logger = req.app.get('stormpathLogger');
 
     function redirectToLoginPage() {
       var url = config.web.login.uri + '?next=' + encodeURIComponent(req.originalUrl);
@@ -30,10 +31,11 @@ module.exports = function (assertGroups, all) {
 
     function isUserInGroups(callback) {
       var isInGroup = false;
-      var done = assertGroups.length;
+      var done = groups.length;
 
       req.user.getGroups(function (err, userGroups) {
         if (err) {
+          logger.info('Could not fetch user ' + req.user.email + '\'s groups.');
           return callback(null, false);
         }
 
@@ -41,7 +43,7 @@ module.exports = function (assertGroups, all) {
         // whether or not it's one of the required groups.
         userGroups.each(
           function (group, iterateNext) {
-            if (assertGroups.indexOf(group.name) > -1) {
+            if (groups.indexOf(group.name) > -1) {
               if (!all || --done === 0) {
                 isInGroup = true;
               }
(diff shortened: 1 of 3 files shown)


lib/middleware/groups-required.js
```diff
@@ -4,63 +4,89 @@ var helpers = require('../helpers');
...
  * Assert that a user is a member of one or more groups before allowing the user
- * to continue.  If the user is not logged in, they will be redirected to the
- * login page.  If the user does not meet the group requirements, they will be
+ * to continue. If the user is not logged in, they will be redirected to the
+ * login page. If the user does not meet the group requirements, they will be
  * shown an unauthorized page letting them know they do not have the required
  * permissions.
  *
- * @param {String[]} groups - A list of groups to assert membership in.  Groups
+ * @param {String[]} assertGroups - List of groups to assert membership for. Groups
```
robertjd: I think `groups` is the better name here, `assertGroups` feels like a boolean parameter.

lib/middleware/groups-required.js
```diff
@@ -4,63 +4,89 @@ var helpers = require('../helpers');
...
+      var done = assertGroups.length;
 
-    req.user.getGroups(function (err, grps) {
-      if (err) {
-        logger.info('Could not fetch user ' + req.user.email + '\'s groups.');
-        return helpers.render(req, res, view);
-      }
+      req.user.getGroups(function (err, userGroups) {
+        if (err) {
+          return callback(null, false);
```
robertjd: This error should be sent to `logger.error()`, so that the developer can know what went wrong.

lib/middleware/groups-required.js
```diff
@@ -4,63 +4,89 @@ var helpers = require('../helpers');
...
+        isUserInGroups(function (err, isInGroup) {
+          if (err || !isInGroup) {
+            res.status(403);
+            return helpers.render(req, res, 'unauthorized');
+          }
+
+          next();
+        });
+      }
+    }, redirectToLoginPage);
```
robertjd: This behavior isn't right, that I'm redirected to the login page if I haven't specified an Accept header.  I should still be able to attempt authorization, regardless of the Accept header value.  It's only when we need to render an error that we should be inspecting the Accept header, to determine how to render the error (JSON error response, or redirect to login page).  And if no Accept value is specified, default to the first value in the produces array.  

I understand that dealing with the first value in the produces array is a larger task for this library, and that it may be out of scope for this PR.  But I still think this PR should be modified to still attempt authorization, regardless of Accept type.

robertjd:
@typerandom this is not passing tests.  Please see inline comments for general code review issues.  We also need to add tests for the new JSON responses.

robertjd:
Thanks @typerandom for the updates.  This is looking good now, but I do want to add a test to assert that the `all` option does assert all groups.  I'm going to bring this back into open, until one of us can add that test.

robertjd:
LGTM, thanks @typerandom for the tests!

Issue #155, opened 2015-10-16
Delete cookies doesn't accept parameters


suryod:
Hi,

When creating accessToken cookie (lib/helpers/create-session.js), it supplies the `accessTokenCookie` parameters. But when deleting accessToken cookie (lib/middleware/delete-cookies.js) it doesn't supplies the parameters. Can you make it so it'll accept the parameters as well. I'm having problem when `accessTokenCookie` is set, users can login but unable to logout because the cookie's domain is not matching.

Thank you,
Indra

typerandom:
Hi Indra.

Thanks for reporting this. We'll be looking into this issue soon. I know that this was reported some time ago, but please, if this still is an issue and is hindering you in your work, please let us know.

Cheers!

harai:
I have exactly the same problem now.

PR #443, opened 2016-06-13
Fix undeletable Cookies

harai:
This fixes undeletable Cookies.

stormpath/express-stormpath#155


robertjd:
Thanks @harai !  I had to open a new PR on this repo, but your patch is included and I've added you to the contributors file :) This patch is in master and will go out in our next release.

PR #453, opened 2016-06-23
Fix registration docs RE name field

### LINKED ISSUE #414: Create Custom fields name issue
Hi,

[Creating Custom Fields](https://docs.stormpath.com/nodejs/express/latest/registration.html#creating-custom-fields) suggests that "**name** - The name of this field, in the form model that is posted to the server." Does this work as documented as the submitted value is not bound to the name value I'm setting in the model when looking formData in preRegistrationHandler's ?

With the following I see the value in preRegistrationHandler to be 'foo,' but I'd expect it to be 'bar'?
```
foo: {
  enabled: true,
  label: 'Test',
  name: 'bar',
  required: true,
  type: 'text'
}
```

Thanks,
Matti

robertjd:
Closes #414

### DIFF
--- a/docs/registration.rst
+++ b/docs/registration.rst
@@ -122,7 +122,6 @@ like this::
             favoriteColor: {
               enabled: true,
               label: 'Favorite Color',
-              name: 'favoriteColor',
               placeholder: 'E.g. Red, Blue',
               required: true,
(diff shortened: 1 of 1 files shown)


typerandom:
Looks good to me!

PR #462, opened 2016-07-06
Fix issue with res.render() in render util not being called with original view path

typerandom:
Fixes so that the render method calls res.render() with the original view path.

#### How to verify

(will be added soon)

Fixes #404


robertjd:
I fixed the test failure on master, we had to update an expected error string.  That change was unrelated to this issue.

PR #465, opened 2016-07-10
Fix session creation for auto-login when resetting password

### LINKED ISSUE #164: autoLogin does not create session for change password flow
This is not creating the password-based token session:

https://github.com/stormpath/express-stormpath/blob/master/lib/controllers/change-password.js#L71

typerandom:
Fixes so that a session cookie is created when `web.changePassword.autoLogin` is enabled and you're resetting the password for an account.

#### How to verify

1. Create a new application using [this gist](https://gist.github.com/typerandom/419d4346aa5e01ade26e50251768cbc9).
2. Start the application.
3. Navigate to the forgot password page (http://localhost:3000/forgot) and reset the password for one of your accounts.
4. Click the reset password link in the email you received and change the password for your account.
5. You should be redirected to `/` and logged in.

Fixes #164

### DIFF
--- a/test/handlers/test-pre-login-handler.js
+++ b/test/handlers/test-pre-login-handler.js
@@ -38,18 +38,21 @@ function preLoginHandlerErrorTestFixture(application, respondWithError, callback
   app.on('stormpath.ready', callback.bind(null, fixture));
 }
 
-function preLoginHandlerInterceptResponseTestFixture(application, respondWith, callback) {
+function preLoginHandlerInterceptResponseTestFixture(application, callback) {
+  var fixture = {
+    expressApp: null,
+    preLoginHandlerFormData: null
+  };
+
   var app = helpers.createStormpathExpressApp({
     application: application,
-    preLoginHandler: function (formData, req, res) {
-      res.json(respondWith);
-      res.end();
+    preLoginHandler: function (formData, req, res, next) {
+      fixture.preLoginHandlerFormData = formData;
+      next();
     }
   });
 
-  var fixture = {
-    expressApp: app
-  };
+  fixture.expressApp = app;
 
   app.on('stormpath.ready', callback.bind(null, fixture));
 }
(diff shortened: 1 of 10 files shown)


robertjd:
I've verified that this PR fixes the reported problem, however I'm also noticing that the pre/post handlers are not being run for this case.  Can we refactor the code to handle those cases as well?  I know it may touch many files, but it seems like we can consolidate a lot of this.

typerandom:
@robertjd The latest changes are now pushed.

test/handlers/test-pre-login-handler.js
```diff
@@ -145,33 +148,29 @@ describe('Pre-Login Handler', function () {
...
+                .set('Accept', 'application/json')
+                .type('json')
+                .send(usernamePasswordBody)
+                .expect(200)
+                .end(function (err) {
+                  if (err) {
+                    return done(err);
+                  }
+
+                  var preLoginHandlerFormData = fixture.preLoginHandlerFormData;
```
robertjd: The intention of this test case is to assert that the developer can end the response inside of their preLoginHandler, and that their response will be the final one.

These modifications don't assert that code path anymore, because the `preLoginHandlerInterceptResponseTestFixture` has been modified to always call `next()`, so we aren't testing a custom response by the developer anymore.  This test should be reverted back to it's original intent.

robertjd:
Thanks @typerandom !  This looks great and I've verified that it solves the original issue.

I do have one issue with a test change, please see my inline comments.  In sum: we need to make sure that we have a test case that asserts the developer's ability to end the response from inside their handler, preventing the rest of the next() handlers from running.  I'd like to address this test issue before we merge.

test/handlers/test-pre-login-handler.js
typerandom: Okay, I see. I'll fix that!

typerandom:
@robertjd I saw your comment and I'll fix the code/test before we continue.

typerandom:
@robertjd This is fixed now.

PR #466, opened 2016-07-10
Fix login/registration handlers for oauth provider login

### LINKED ISSUE #451: Post-registration handlers not called during social login w/ access token flow
If you are posting provider data to the `/login` endpoint, to log in a user with an access token that was obtained from a social provider, the code routes you through `helpers.loginWithOAuthProvider()`, which will run post-login handlers, but not post-registration handlers.  The "created" status of the account can be determined by looking at the `providerAccountResult.created` property, and should be used to fork between login and registration flows.

### LINKED ISSUE #452: Pre-login & registration handlers not called during social login w/ access token flow
When posting provider data to the /login endpoint, to log in a user with an access token that was obtained from a social provider, we are not running the pre-login handlers because we fork to `helpers. loginWithOAuthProvider ()` and bypass `handlePreLogin`, seen here:

https://github.com/stormpath/express-stormpath/blob/master/lib/controllers/login.js#L97-L101

That said, we also need to fork for login vs registration cases, similar to the issue described in #451

typerandom:
Fixes so that login/registration pre/post handlers are triggered when authenticating with an OAuth provider.

#### How to verify

Create a sample application that defines pre and post login, registration handlers.

When using the access token flow for a social provider, and authenticating for the first time (and account will be created in stormpath), the following handlers should be run:

* preRegistrationHandler
* postRegistrationHandler
* preLoginHandler
* postLoginHandler

Then when authenticating for a second time (the account already exists in stormpath), the following handlers should run:

* preLoginHandler
* postLoginHandler

Fixes #451 
Fixes #452

### DIFF
--- a/lib/helpers/login-with-oauth-provider.js
+++ b/lib/helpers/login-with-oauth-provider.js
@@ -51,7 +51,7 @@ module.exports = function loginWithOAuthProvider(options, req, res) {
       });
     }
 
-    function continueWithHandlers(preHandler, postHandler) {
+    function continueWithHandlers(preHandler, postHandler, onCompleted) {
       preHandler(options, req, res, function (err) {
         if (err) {
           return oauth.errorResponder(req, res, err);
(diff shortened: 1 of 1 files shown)


robertjd:
Thanks @typerandom ! 

I was just testing this with the [Stormpath Angular + Express Fullstack Sample Project](https://github.com/stormpath/express-stormpath-angular-sample-project).  I've edited the top of this PR with what I'm looking for as I test manually.

I believe we're missing one case here: I don't see the `preLoginHandler` being run during the registration flow, and I think it should be.

typerandom:
@robertjd Login handlers will now run during registration.

robertjd:
I'm seeing the `postLoginHandler` being called twice, @typerandom can you confirm?

typerandom:
@robertjd Yes, sorry about that. It's actually been fixed in refactor that was done in #465. See https://github.com/stormpath/express-stormpath/pull/465/files#diff-260b00911d1b53968e79acacf8308d51L50. So we should probably have that merge in before.

robertjd:
@typerandom in that case can you please rebase this against #465 ?

typerandom:
@robertjd The changes from #465 that we were depending on are now applied in this PR.

robertjd:
@typerandom do you know what's causing the test failures?

typerandom:
@robertjd 

I don't. But I'll look into what is failing!

typerandom:
This is failing because some of the refactored code from #465 that I added. I thought it was possible to just include it, but it turned out there were other parts that were dependent. These issues are fixed in #465, so we need to merge that in, then do a rebase before merging this.

typerandom:
@robertjd This has now been rebased against master.

PR #474, opened 2016-07-21
Surface token exchange errors via info() logger

robertjd:
This will make it easier to debug issues with social login flows, where things can go wrong at this step, specifically if the clock of the local sever is not in sync with ntp, and we get a token expired error because of it.


timothyej:
Code looks good but I haven't tried it myself.

typerandom:
This is still in the `In Progress` column. Should this be reviewed/tested or do you want to add something more?

robertjd:
Oops, should have been in review.

typerandom:
Looks good! I'm merging this in.

PR #496, opened 2016-08-08
Fixes for social auth code flow

robertjd:
@edjiang I'd like you to take a look at these changes

In it's current state, this library constructs the "redirect url" value (used in several places by several providers) by looking at the incoming value of the Host header, when the user is being redirected back to this library, from the social provider.

This caused a problem because Express.js doesn't provide the port number when parsing the Host header.  See https://github.com/expressjs/express/issues/2179 for a description of the problem.  As such, the redirect URL that is configured on your directory may be different than the one that is dynamically evaluated by this library, because the hostname is not correct, thus causing a validation error when exchanging the code for an access token.

The quick solution was to write a helper method that doesn't have the problem of stripping the port from the header.  But I've taken it a step further, and just started pulling the value from the directory configuration, for providers that have a configured redirect URL.

I also made a partial refactor: we were doing the code exchange in this library for Linkedin, but we don't need to, the stormpath API will do that for you.  A such I made a not-DRY copy of the Google login handler, into the handler for linkedin.  These could be refactored into a common handler.

**To Reproduce The Problem**

Grab the [https://github.com/stormpath/express-stormpath-sample-project](https://github.com/stormpath/express-stormpath-sample-project) and configure a Google, Linkedin or Github directory.  Set the redirect URL to http://localhost:8080.  Configure nginx or some other proxy to proxy 8080 to 3000, and run the sample project on port 3000.  When you try to complete the social login flow, you should get an error about the redirect URI being mismatched.


edjiang:
Hey @robertjd ,

Just curious, did you want me to UAT the nginx thing? I can try testing to make sure that it works if you'd like =]

In terms of the general approach, I think this is fine for now, but one thing to keep in mind is that in the future we may allow multiple redirect URIs for each social provider directory. This is so they can redirect back into mobile, web, staging, and potentially use the same directory across multiple applications. 

So -- once Social Login V2 comes out, we might have to change up the approach here again. 

I've been modeling an Authorization Code as a pair of `code` and `redirectURI` in Turnstile so far, as many OAuth providers will verify that the code is submitted to the right `redirectURI` as well.

robertjd:
Thanks, it's the general feedback I was really looking for.  I'm also concerned that tying to a specific callback URI is limiting, but since we only allow one on the directory right now I'm not actually introducing a new limitation, so I'm okay with this for now as it does reduce some problem area.

In a world where we allow more than one redirect URL to be specified, how does the web application know which URL to use when doing a code exchange?

edjiang:
Sounds good!

When we allow multiple redirect URIs, it would most likely have to be the application reading from the request host, scheme, and path. So actually your "quick solution" would probably be a better solution in the long run. 

EG: Google redirects a user to https://example.com/callbacks/google?code=XYZ

The access token exchange requires:

```
client_id
redirect_uri
code
```

(`client_secret` is often required by providers but not required by the spec)

Then the application will submit the three pieces of info and get an access token. Without a matching `redirect_uri` set, providers will usually deny the exchange and still consume the authorization code to prevent further use.

robertjd:
Gotcha, thanks for the feedback.  I'm going to simplify this fix and just go for the dynamic host resolution fix, I'm closing this PR in favor of #498

PR #499, opened 2016-08-10
Release 3.1.4

robertjd:


typerandom:
@robertjd This looks good! 👍  I still have a bit to install on my computer so if you can publish this, that would be great!

PR #506, opened 2016-08-18
Login docs should use `login` instead of `username`

### LINKED ISSUE #486: Wrong docs for /login POST
https://docs.stormpath.com/nodejs/express/latest/login.html#json-login-api

It says username/password should be in the JSON object, but it should be 'login' and password.

typerandom:
Fixes the login docs so that `login` is used instead of `username`.

#### How to verify

Review the doc changes.

Fixes #486

### DIFF
--- a/docs/login.rst
+++ b/docs/login.rst
@@ -19,7 +19,7 @@ create the proper session cookies.
 Next URI
 --------
 
-The form will render with two fields for username and password, and this form
+The form will render with two fields for login and password, and this form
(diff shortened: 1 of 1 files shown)


docs/login.rst
```diff
@@ -47,7 +47,7 @@ field configuration:
         form: {
           fields: {
             login: {
-              label: 'Your Username or Password',
```
robertjd: lulz

PR #508, opened 2016-08-24
Fix: Social login not respecting ?next

### LINKED ISSUE #482: Social logins are not being redirected properly
Issue reported by one of our users:

**When a user navigates to an internal page on my site, but is not logged in, I expect him to be taken to the login page and then be redirected to the originally-requested internal page upon successful login.**

It looks like this is also the intention of the Stormpath code, but it is not working.

I logout. 
I navigate to an internal page. I'm redirected to https://localhost/api/v1/sdriver/login?next=%2Fapi%2Fv1%2Fdriver%2F

Looks good. I'm taken to the login page and there is a next param pointing to the desired target.

I then do a **social login**, via the Google button on my login page. 
**After login, I'm taken to my Stormpath web.login.nextUri page, and not to the desired target**.

Looking at the express-stormpath code, it appears that it is trying to handle this case: 
https://github.com/stormpath/express-stormpath/blob/master/lib/controllers/google-login.js#L67 checks first for query.next, and only then for nextUri.

But, problem. By the time the code gets there, it has lost the next field.

```
req.url is:"/callbacks/google?code=4/Je1WzWBbiKTkDHwsM6Bv0vyVc9W2ClPVVPMajZrU8uo&scope=https://www.googleapis.com/auth/userinfo.email+https://www.googleapis.com/auth/userinfo.profile"
```

and req.query only holds: 

```
{ 
code: "4/Je1WzWBbiKTkDHwsM6Bv0vyVc9W2ClPVVPMajZrU8uo", 
scope: "https://www.googleapis.com/auth/userinfo.email https://www.googleapis.com/auth/userinfo.profile" 
}
```

I don't know if this is a general bug, or not supported yet, or caused by some interaction with my nginx-proxied config, or some interaction with my use of social-login unification (copied from https://github.com/stormpath/stormpath-express-social-unification-example)

typerandom:
Fixes so that login with social providers support the `?next` query string parameter.

#### How to verify

1. Checkout this branch.
2. Run `$ npm link`.
3. Use [this](https://gist.github.com/typerandom/0887f0085c77ccadd2f04859ce876385) example application.
4. Link it to the branch `$ npm link express-stormpath`.
5. Run the application.
6. Navigate to `http://localhost:3000/hello`.
7. Verify that you are redirected to `http://localhost:3000/login?next=%2Fhello`.
8. Login with either Facebook, Google, GitHub or LinkedIn.
9. Verify that you are redirected back to `http://localhost:3000/hello`.

#### Discussion

The API-key secret is used to sign the `oauthStateRedirectUri` cookie. Any objections to this?

Fixes #482

### DIFF
--- a/lib/controllers/facebook-login.js
+++ b/lib/controllers/facebook-login.js
@@ -75,7 +75,11 @@ module.exports = function (req, res) {
             return oauth.errorResponder(req, res, err);
           }
 
-          var nextUrl = url.parse(req.query.next || '').path || (resp.created ? config.web.register.nextUri : config.web.login.nextUri);
+          var nextUrl = oauth.common.consumeRedirectUri(req, res);
+
+          if (!nextUrl) {
+            nextUrl = url.parse(req.query.next || '').path || (resp.created ? config.web.register.nextUri : config.web.login.nextUri);
+          }
 
           if (resp.created && registrationHandler) {
             registrationHandler(req.user, req, res, function () {
--- a/lib/controllers/facebook-login.js
+++ b/lib/controllers/facebook-login.js
@@ -78,7 +78,7 @@ module.exports = function (req, res) {
           var nextUrl = oauth.common.consumeRedirectUri(req, res);
 
           if (!nextUrl) {
-            nextUrl = url.parse(req.query.next || '').path || (resp.created ? config.web.register.nextUri : config.web.login.nextUri);
+            nextUrl = resp.created ? config.web.register.nextUri : config.web.login.nextUri;
           }
 
           if (resp.created && registrationHandler) {
--- a/lib/controllers/facebook-login.js
+++ b/lib/controllers/facebook-login.js
@@ -1,7 +1,5 @@
 'use strict';
 
-var url = require('url');
-
 var helpers = require('../helpers');
 var oauth = require('../oauth');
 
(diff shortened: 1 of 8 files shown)


lib/controllers/facebook-login.js
```diff
@@ -75,7 +75,11 @@ module.exports = function (req, res) {
             return oauth.errorResponder(req, res, err);
           }
 
-          var nextUrl = url.parse(req.query.next || '').path || (resp.created ? config.web.register.nextUri : config.web.login.nextUri);
+          var nextUrl = oauth.common.consumeRedirectUri(req, res);
+
+          if (!nextUrl) {
+            nextUrl = url.parse(req.query.next || '').path || (resp.created ? config.web.register.nextUri : config.web.login.nextUri);
```
robertjd: Do we need to keep this: ` url.parse(req.query.next || '').path` ? It seems that this will never be available during a social login flow, and that is why we're doing the cookie workaround?

robertjd:
@typerandom I've reviewed this and it looks good, though I do have one question about some code that I think can be removed, please see my inline comment.

lib/controllers/facebook-login.js
typerandom: That is a valid point and I think you're right about it. I'll probably just remove it.

typerandom:
@robertjd I have now removed `url.parse(req.query.next || '').path`. Please review again :)

Issue #507, opened 2016-08-23
postLoginHandler should work with OAuth2 Password Grant(?)


passatgt:
I'm not 100% sure, but i think postLoginHandler should run when a new access token is generated using /oauth/token, especially with grant_type=password, which is like a login function on mobile clients.

robertjd:
Hi @passatgt , yes the `/oauth/token` endpoint should trigger the post-login handler if tokens are issued.  I confirmed that this isn't working and I'm going to open this as a bug.  Do you have a workaround in the meantime?  Let me know if I can help you find one, as that may be quicker (we're working on a few other issues this week and I'm not sure how quickly we can get to this one).  Thanks for the report!

passatgt:
Currently i just setup another endpoint which is called by the app after successful login. I did not find a way to hook up a middleware to the oauth/token route after the token was returned, but if you have another idea, that would be great.

PR #510, opened 2016-08-24
postLoginHandler for OAuth2 Password Grant

passatgt:
Fix for issue #507


robertjd:
Thanks @passatgt - though I don't think we want this to be triggered for the `refresh_token` workflow, only `password`.

passatgt:
Ok, makes sense, i guess it works like this, i just didn't wanted to duplicate the case, because otherwise its the same.

robertjd:
Can we update this to use the existing switch structure?  Thanks!

passatgt:
I had to change it yet again. The account is not returned by the request, added client.getAccount instead.

PR #521, opened 2016-09-06
Add Authorization Code and Access Token docs for social login

### LINKED ISSUE #380: Document social login, access token vs. authorization code
We support both, so we should update the documentation to describe how each can by done, and a little information on why you might choose one over the other.

typerandom:
Add Authorization Code and Access Token docs for social login.

#### How to verify

1. Run `$ npm run docs`.
2. Open `/docs/_build/html/social_login.html` to view the new docs.

#### Discussion

* Part of the issue discussion was adding a note about using `postmessage` as the callback uri when authenticating using the Google provider. But I did not really understand that, and the Google docs about `postmessage` was very thin.

* I'm not happy about the placement of the new docs. The sections `Logging in using an Authorization Code` and `Logging in using an Access Token` are currently placed after the section `Test it Out`. This feels strange. Any suggestions on how to make it better?

Fixes #380

### DIFF
--- a/docs/social_login.rst
+++ b/docs/social_login.rst
@@ -111,7 +111,46 @@ prompted to accept the permissions requested:
 After accepting permissions, you'll be immediately redirected back to your
 website at the URL specified by ``redirectUrl`` in your app's config.
 
+.. note::
+
+  By default the configuration ``web.social.facebook.scope`` is set to ``email``. Change
+  this configuration value to request additional scopes.
+
 Simple, right?!
+
+Logging in using an Access Token
+................................
+
+If you want to use a Facebook SDK to login a user you can easily do so by grabbing the
+Access Token from the SDK login response and then POST it to the ``/login`` endpoint.
+
+To show you exactly what that means, let's say that you're using the JavaScript SDK and
+calling ``FB.login(...)``:
+
+.. code-block:: javascript
+
+  FB.login(function (response) {
+    if (!response.authResponse) {
+      return console.error('User cancelled login or did not fully authorize.');
+    }
+
+    // Login by making POST request to /login with the value from:
+    // response.authResponse.accessToken
+  });
+
+Now, using the Access Token received in the callback response (``response.authResponse.accessToken``) you can create a new ``POST`` request to the
+``/login`` endpoint with a JSON body as shown below:
+
+.. code-block:: javascript
+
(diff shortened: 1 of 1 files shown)


docs/social_login.rst
```diff
@@ -111,8 +111,47 @@ prompted to accept the permissions requested:
...
 website at the URL specified by ``redirectUrl`` in your app's config.
 
+.. note::
+
+  By default the configuration ``web.social.facebook.scope`` is set to ``email``. Change
+  this configuration value to request additional scopes.
+
 Simple, right?!
 
+Logging in using an Access Token
```
mdeggies: This looks good, but I see what you mean about the title & placement of the new section. I think this section and ‘Test it Out’ should be renamed. I would rename ‘Test it Out’ to something like Standard auth, or Default auth. This implies that the standard/easy way of doing things is to visit /login, click on the FB button, input credentials, etc. Then you can rename your section to something like Facebook SDK Auth, or Facebook Access Token Auth, since this section shows an alternative. Same with Google/LinkedIn sections

docs/social_login.rst
```diff
@@ -111,8 +111,47 @@ prompted to accept the permissions requested:
...
+................................
+
+If you want to use a Facebook SDK to login a user you can easily do so by grabbing the
```
mdeggies: I really like your explanations at the beginning of the Google & LinkedIn sections. I would add that same explanation here if it applies. "Using the Access Token (implicit) workflow is preferable when you're logging in from a place where the storage can be trusted. E.g. from your back-end where only you have access to the Access Token."

oshalygin:
Really good stuff, directly relates to what I'm currently trying to accomplish with Google.  With that, the redirect URI is something I need to handle at the React level.  I get the response back which includes the queryString params of `state` and `code`.

Is this something specific to the individual FE SDK and thus should be excluded from the docs since they're really meant for server side applications which are agnostic to the underlying SDK?

My redirect looks something like this(Note the `http://localhost:8080/authorize/callback` string is configurable at the GCP console, the rest is what you get back after you've authenticated with Google):

http://localhost:8080/authorize/callback?state=df81b8ae-a25b-781a-b45f-832&code=4/Jtq8_nbU6A2vvixT82Bsak73gCigvnebsE

PR #524, opened 2016-09-12
Moving getUser into middleware folder for better organization.

rdegges:
This does exactly what the title says it does. All tests pass locally.


typerandom:
@rdegges Thank you man. Makes sense! :) I did a rebase and ran tests locally too. Works as expected so going to merge this in.

PR #533, opened 2016-10-04
Fix confusing setup docs

### LINKED ISSUE #490: Confusing Docs / Behavior for Applications
The quickstart docs (https://docs.stormpath.com/nodejs/express/latest/setup.html#initialize-express-stormpath) suggest that all configuration is optional so long as the API key file is available as `~/.stormpath/apiKey.properties`.

I've got a Stormpath tenant with a ton of Applications in it, and this does work without having me specify an Application, but this seems really confusing. I think the desired behavior should be that we throw an error if there are more than 2 applications, and force the developer to specify the application they want to use.

We should also update the documentation in the quickstart linked above to talk about this.

typerandom:
Fix confusing setup docs.
#### How to verify

Check the changes and verify that it solves the problem described in the issue.

Fix #490


mdeggies:
Hey Robin! Since we're asked about this stuff **a lot**, I added some more examples and explanations to the configuration page. The structure is very similar to Nate's asp.net core config page, which I really like: https://docs.stormpath.com/dotnet/aspnetcore/latest/configuration.html

I thought it would be better to push here rather than submitting a new PR, but LMK

typerandom:
I'm happy with these changes! Merging in.

PR #550, opened 2016-11-25
Scope factory and client_credentials

### LINKED ISSUE #535: Client credentials returns incorrect access token TTL
Making a request to my local express server's /oauth/token endpoint always returns an access token TTL of 3600, regardless of my Application's access token TTL settings.

This

`curl -X POST --user $ACCOUNT_ID:$ACCOUNT_SECRET http://localhost:3000/oauth/token -d grant_type=client_credentials`

returns

`{"access_token":"eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiI2WFU5UzhLU0hHR1lTVVlZNVRaNTNVWkJIIiwiaXNzIjoiaHR0cHM6Ly9hcGkuc3Rvcm1wYXRoLmNvbS92.....","token_type":"bearer","expires_in":3600,"scope":""}`

Making the same request to the https://api.stormpath.com/v1/applications/$APP_ID/oauth/token?grant_type=client_credentials&client_id=$USER_ID&client_secret=$USER_SECRET results in the correct expires_in time.

Tweety-FER:
Integrates new Node SDK `client_credentials` authenticator and adds support for scope factories.

Depends on https://github.com/stormpath/stormpath-sdk-node/pull/575 PR in the Node SDK and will not work without it. This PR allows the `OAuthAuthenticator` class to proxy the scope factory and signing key between the Express app and the concrete authenticator instances.

The scope factory is set as a property in the config (`config.web.scopeFactory`), and will be used if defined there. The client secret will be used automatically, and no configuration is required.

It adds support for sending the client id and secret as parameters, but remains backwards-compatible with sending them via the `Authorization` header.

**Note**: There still seems to be some sort of an issue with the code, which I am trying to resolve. When using `client_credentials` in the test, it will *sometimes* fail unless it is awaited for a short time between creating the api keys and using them. Still not sure why this is happening, but am looking into it. Could always use a second pair of eyes. At the moment, I'm trying to solve it in a hacky way, using a `setTimeout`.

Fixes #535


Tweety-FER:
The guys who worked on the Ruby and Rails SDKs tell me that this has been known to happen - elastic search syncing or something like that. Don't know if that's a problem. I do assume that it's extremely unlikely that the exact conditions from the test will be replicated, though - even delaying with something as simple as a `console.log` will make the error not appear, and I doubt someone is creating and using the client credentials in the same breath in any real-world application. Maybe we don't have a problem there after all.

PR #552, opened 2016-11-29
Add missing parenthesis to user data example

dominicliu:
In the example there is a missing closing parenthesis.


robertjd:
Thanks @dominicliu !

Issue #542, opened 2016-11-08
Regression in 3.1.6: Error cookieParser("secret") required for signed cookies


ppitonak:
1. upgrade express-stormpath to 3.1.6 (from 3.1.5)
2. navigate to https://example.com/login

This error is displayed/logged: 

```
Error: cookieParser("secret") required for signed cookies
   at ServerResponse.res.cookie (/home/ubuntu/workspace/node_modules/express/lib/response.js:785:11)
   at Object.module.exports [as setTempCookie] (/home/ubuntu/workspace/node_modules/express-stormpath/lib/helpers/set-temp-cookie.js:20:7)
   at text/html (/home/ubuntu/workspace/node_modules/express-stormpath/lib/controllers/login.js:94:15)
   at Object.handleAcceptRequest (/home/ubuntu/workspace/node_modules/express-stormpath/lib/helpers/handle-accept-request.js:59:3)
   at module.exports (/home/ubuntu/workspace/node_modules/express-stormpath/lib/controllers/login.js:27:11)
   at Layer.handle [as handle_request] (/home/ubuntu/workspace/node_modules/express/lib/router/layer.js:95:5)
   at next (/home/ubuntu/workspace/node_modules/express/lib/router/route.js:131:13)
   at forceDefaultBodyMiddleware (/home/ubuntu/workspace/node_modules/express-stormpath/lib/helpers/body-parser.js:41:7)
   at Layer.handle [as handle_request] (/home/ubuntu/workspace/node_modules/express/lib/router/layer.js:95:5)
   at next (/home/ubuntu/workspace/node_modules/express/lib/router/route.js:131:13)
   at Route.dispatch (/home/ubuntu/workspace/node_modules/express/lib/router/route.js:112:3)
   at Layer.handle [as handle_request] (/home/ubuntu/workspace/node_modules/express/lib/router/layer.js:95:5)
   at /home/ubuntu/workspace/node_modules/express/lib/router/index.js:277:22
   at Function.process_params (/home/ubuntu/workspace/node_modules/express/lib/router/index.js:330:12)
   at next (/home/ubuntu/workspace/node_modules/express/lib/router/index.js:271:10)
   at /home/ubuntu/workspace/node_modules/express-stormpath/lib/middleware/get-user.js:173:16
   at Application.authenticateApiRequest (/home/ubuntu/workspace/node_modules/stormpath/lib/resource/Application.js:1094:12)
[...]

PR #553, opened 2016-11-29
Revert usage of signed cookies for oauth state token

### LINKED ISSUE #551: Regression: LinkedIn login broken in 3.1.6
Seems like we switched to using signed cookies in 3.1.6 for the OAuth State token, but are not decoding them as signed cookies. So the OAuth state doesn't match, and thus gets rejected with `Invalid state token provided`

Two questions:

1. Why'd we switch to signed cookies? I'm not sure if I see the benefit here.
2. We need to be decoding these cookies as signed cookies, or revert this change, as currently LinkedIn (and I'm guessing GitHub) login is broken with the `authorization_code` flow.

edjiang:
Fixes #551 #542

### DIFF
--- a/lib/controllers/facebook-login.js
+++ b/lib/controllers/facebook-login.js
@@ -96,7 +96,7 @@ module.exports = function (req, res) {
   }
 
   if (req.query.code) {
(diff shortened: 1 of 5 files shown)


robertjd:
Thanks @edjiang as discussed i'm modding this pr to remove the cookie signing strategy entirely, as it will also reduce the friction that we saw in #542

PR #557, opened 2016-12-13
Saml integration

### LINKED ISSUE #492: SAML buttons don't appear
After mapping a SAML directory to my Application, and spinning up the express project, no SAML buttons appear on the login page. Without creating your own button and route to kickoff the flow, you have to use ID Site.

Tweety-FER:
Implements the SAML flow for traditional websites. The URL is built on-demand (to keep the timestamps current in the JWT) and the user is redirected to the provider, through the usual flow, and back to the configured URL, which then verifies the token and redirects the user appropriately.

All of this happens automatically if `web.saml.enabled` is set to true in the configuration.

As for the SPA flow, it currently only tosses the provider data in the login form view model, which is the old behaviour. This does not include the idp url generation. @robertjd Should it? To preserve the same behaviour as the traditional websites, it would have to get a fresh token with the redirect URI. 

The only idea that comes to mind without adding verification logic to the SPAs is to have the SPA also redirect the user to the `/verify` page, passing a query parameter so that the `/saml-verify` endpoint knows to redirect the user back to the SPA. 

Alternatively, it could redirect back to the app and have it exchange it via `stormpath_token` grant, like the Client API social flow. In this case, all express would have to do is generate a fresh token in the initial redirect URL. Not sure what you intend to do with the Client API changes, as well.

This part is still a WIP, obviously.

Also, we talked about route naming, so that's something that ought to be checked.

Fixes #492 (once the SPA part is done)


lib/controllers/saml-redirect.js
```diff
@@ -0,0 +1,40 @@
...
+ * @method
+ *
+ * @param {Object} req - The http request.
+ * @param {Object} res - The http response.
+ */
+module.exports = function (req, res) {
+  var application = req.app.get('stormpathApplication');
+  var builder = new stormpath.SamlIdpUrlBuilder(application);
+  var config = req.app.get('stormpathConfig');
+  var cbUri = req.protocol + '://' + req.get('host') + config.web.saml.verifyUri;
```
mheisig: I'm using a similar approach to support SAML login. I'd recommend support a `host` config option here instead of assuming the `req.host`. Running this in a Docker container behind an nginx proxy winds up with the `host` being reported as the Docker service name instead of the public URI.

PR #580, opened 2017-01-20
Update stormpath config dependency

### LINKED ISSUE #579: Cannot read property getGroups of undefined
Essentially we need to update `stormpath-config` to `0.0.27` but currently its at `0.0.26`.  Maybe consider always updating the patch versions if the plan is to follow semver?

I'll make a PR shortly.

Links to:
https://github.com/stormpath/stormpath-sdk-node/issues/596

oshalygin:
- The dependency stormpath-config handles the configuration
  helpers that are used by the SDK.
- This update is necessary as there is an update within the
  EnrichIntegrationFromRemoteConfigStrategy file which
  sets the client on the outerScope object.  Without this
  the calls to getGroups wil yield undefined.  This was
  resolved with this commit:
  https://github.com/stormpath/stormpath-node-config/commit/49461660d2614cbf130cb788f458fd5e0cd1c996

closes #579

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -6,6 +6,15 @@ Change Log
 
 All library changes, in descending order.
 
+Version 3.1.9
+-------------
+
+**Released January 20, 2017.**
+
+- The following production dependencies have been updated:
+
+  - ``stormpath-config@0.0.26`` -> ``stormpath-config@0.0.27``
+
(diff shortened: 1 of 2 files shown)


oshalygin:
This PR assumes you can(hopefully) push the changes out to npmjs :) 👍

PR #584, opened 2017-01-23
Fix field validation

Tweety-FER:
Fixes several issues with the account field validation helper:
1. Error generation used `field.label || field.label` instead of `field.label || field.name`
2. It considered all falsy values to be missing when `required: true`, when `false` or `null` can be legitimate values.
3. It did not validate required custom fields directly. It now looks into `customData`

Fixes https://github.com/stormpath/stormpath-express-react-example/issues/29


robertjd:
Thanks @Tweety-FER !  I've added a test for one more case, LGTM.

PR #585, opened 2017-01-24
Use new access token authenticator (Client API Support)

robertjd:
**Work In Progress**

Updates this library to use the new `StormpathAccessTokenAuthenticator`, in a backwards-compatible way, so that we can release a patch release that will allow this library to properly authenticate access tokens that are issued by the Client API.

Depends on the `access-token-authenticator` branch in the Node SDK, as seen in this PR:

https://github.com/stormpath/stormpath-sdk-node/pull/597

Todo:

- [x] Update authentication and authorization docs to discuss Client API and be clearer about how to use our access tokens.

### DIFF
--- a/docs/authentication.rst
+++ b/docs/authentication.rst
@@ -3,75 +3,82 @@
 Authentication
 ==============
 
-This library offers several options for securing your application and
-authenticating your users.  Which strategy should you use?  The answer depends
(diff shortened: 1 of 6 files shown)


typerandom:
I made some formatting changes, but other than that, I'm 👍  for this!

Issue #586, opened 2017-01-25
Add idSiteResult explanation to docs 


mdeggies:
Need to add an idSiteResult explanation to the express docs that explains how/when to use it.

mdeggies:
https://github.com/stormpath/express-stormpath/pull/587

PR #587, opened 2017-01-25
#586: /idSiteResult endpoint

mdeggies:
Added a little more info about the /idSiteResult endpoint to explain what it's used for and how it can be used in the SAML/Social login workflows. I think this will work for now until the stormpath callback endpoint is implemented and documented.

### DIFF
--- a/docs/login.rst
+++ b/docs/login.rst
@@ -247,9 +247,9 @@ Then you can enable ID Site in your express configuration::
 When ID Site is enabled, any request to ``/login`` or ``/register`` will cause a
 redirect to ID Site.  When the user is finished at ID Site, the user will be
 redirected to the ``/idSiteResult`` endpoint on your application with a secure JWT
-passed as a URL parameter. The job of ``/idSiteResult`` is to validate this JWT
-and return the Stormpath Account. After this is done, the authenticated user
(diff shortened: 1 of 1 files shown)


docs/login.rst
```diff
@@ -245,10 +244,17 @@ Then you want to enable ID Site in your express configuration::
...
-When ID Site is enabled, any request for ``/login`` or ``/register`` will cause a
-redirect to ID Site.  When the user is finished at ID Site they will be
-redirected to `/idSiteResult` on your application.  Our library will handle
-this request, and then redirect the user to the ``nextUri``.
-
+When ID Site is enabled, any request to ``/login`` or ``/register`` will cause a
+redirect to ID Site.  When the user is finished at ID Site, the user will be
+redirected to the ``/idSiteResult`` endpoint on your application with a secure JWT
+passed as a URL parameter. The job of ``/idSiteResult`` is to validate this JWT
+and return the Stormpath Account. After this is done, the authenticated user
```
robertjd: It doesn't return the account directly, I'd like us to cover: it exchanges the jwtResult for an access and refresh token, which are stored in cookies during the redirect to the `nextUri`

robertjd:
Thanks @mdeggies !

PR #603, opened 2017-02-21
Update protected route middleware docs

### LINKED ISSUE #602: Cookie parser is required for strompath.authenticationRequired
Came across this in my testing earlier but if you protect a route with `stormpath.authenticationRequired`, the authentication check will not occur if the express application didn't bring in the `cookie-parser` middleware via the following:

```javascript

import express from 'express';
import stormpath from 'express-stormpath'
import cookieParser from 'cookie-parser'

const app = express();
app.use(cookieParser());

application.use(stormpath.init(application, stormpathConfiguration));

// This will always respond with authenticated(well reroute) if cookie-parser wasn't included.
application.get('/myAwesomeRoute', stormpath.authenticationRequired, function(req, res){ 
    res.status(200).json(someAwesomeObject);
}
```

Proposing to add this to the `README.md` as a required component right above this section:
**9. Protect Your Routes**

Open to making a PR if you're comfortable with the addition @robertjd

oshalygin:
- The stormpath middleware, _stormpath.authenticationRequired_, depends
  on cookie-parser being defined.  Cookie-parser will attach cookies
  from the request and attach them to the canonical req object in
  express.  This property is then used by the middelware to verify
  the authenticity of the user and properly secure the route.

- This commit provides additional requirements to users who are
  leveraging the authenticationRequired middleware

Closes #602

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,25 @@ Follow these steps to add Stormpath user authentication to your Express.js app.
     //...
   });
   ```
+  
+  The `stormpath.authenticationRequired` middleware depends on the `cookie-parser` middleware, make sure that you are including the `cookie-parser` prior to all of your secured routes:
+
+  ```javascript
+  var express = require('express');
+  var stormpath = require('express-stormpath');
+  var cookieParser = require('cookie-parser');
+
+  var app = express();
+
+  // Include the cookier-parser middleware prior to securing the route with 'stormpath.authenticationRequired'
+  app.use(cookieParser());
+  app.use(stormpath.init(application, stormpathConfiguration));
+
+
+  app.get('/secret', stormpath.authenticationRequired, function(req, res){ 
+    //...
+  });
+  ```
 
   For API services that use HTTP Basic Auth, use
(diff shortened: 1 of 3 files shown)


oshalygin:
This will never be merged :( @robertdamphousse-okta @robertjd it was good run man!

oshalygin:
we should close it

robertjd:
Hi @oshalygin , yes I will close this one, thanks.  Internally this library requires cookie-parser, and uses it ([see here](https://github.com/stormpath/express-stormpath/blob/master/lib/stormpath.js#L147)), so it shouldn't be necessary to add it in?  If you've found otherwise I'd love to see the repo.

oshalygin:
@robertjd Yeah it's required, I'm kind of tied up right now but I'll put together an example repo :) It's necessary because the cookie needs to be read first from the request and then passed to the SDK middleware.  Otherwise, you'll always get unauthenticated if you try to use the middleware

`router.use(stormpath.authenticationRequired);`

PR #608, opened 2017-03-09
Refactor authentication to use Okta authorization server

robertjd:
Authentication now happens against an Okta Authentication Server

Token validation is done with the Okta access token, and the application users API.

Logout uses revoke on the Authorization Server.

Begin writing the changelog for this next major version.

Tests will be failing in this PR, as I have not yet addressed the functional differences, or the change to ES6 classes.

This branch still requires you to provide Stormpath configuration somewhere, as I haven't yet removed those dependencies.

Todos:
- [x] Example configuration block in changelog should not use literal values, use example values e.g. `https://dev-<my org id>.oktapreview.com`
- [x] Remove ES6 code, we've decided not to force a Node upgrade during the transition.
- [x] The authorization server configuration can be discovered through the API, so we should do this so that the developer only needs to specify their API token and app ID.
- [ ] More help around creating an application and authorization server.  I want to create a screencast where I walk through this, as it's a lot of moving parts.
- [x] New authentication result from Okta oauth endpoint needs to match `OAuthPasswordGrantAuthenticationResult` as much as posssible
- [x] Normalize base URL trailing slash
- [x] Attempt to get some of the IT tests running


mraible:
What do I need to change in this file to make this work with Okta? 

https://github.com/stormpath/stormpath-angular2-express-example/blob/master/server/server.js

Related: how do I setup my Okta instance to work with this?

robertjd:
Thanks @mraible , great questions.  Take a look at the mods to the changelog?  If it's not clear let me know and I'll get more detailed.  As we go through this migration I want to use the changelog to collect all the notes.

docs/changelog.rst
```diff
@@ -6,6 +6,57 @@ Change Log
...
+  the Application is a Client for your Authorization Server.
+- **Org**:  In Stormpath you had a Tenant, and in Okta you have an Org.  Every
+  Org has it's own distinct URL.
+
+These new properties should be provided like so:
+
+.. code-block:: javascript
+  app.use(stormpath.init(app, {
+    okta: {
+      org: 'https://dev-613050.oktapreview.com/',
```
rchild-okta: nit: Looks like below we're expecting this to not have the / at the end

lib/middleware/get-user.js
```diff
@@ -112,48 +114,21 @@ module.exports = function (req, res, next) {
...
-
-      authenticationResult.getAccount(function (err, account) {
+      getAppUserById(config.okta.org, config.okta.appId, config.okta.apiToken, jwt.body.uid, (err, user) => {
         if (err) {
-          logger.info('Failed to get account ' + authenticationResult.account.href);
+          logger.info(err);
           return next();
         }
-
-        if (account.status !== 'ENABLED') {
```
rchild-okta: Double checking - in Okta, if a user is not active, it'll return an error right (i.e. so we don't need to keep a check like this)?

lib/okta/access-token-authenticator.js
```diff
@@ -0,0 +1,87 @@
...
+    return this;
+  }
+
+  forOrg(org) {
+    this.org = org;
+    return this;
+  }
+
+  withClientId(clientId) {
+    this.clientId = clientId;
```
rchild-okta: Is this supposed to also be a fluid function (i.e. return this)?

lib/okta/access-token-authenticator.js
```diff
@@ -0,0 +1,87 @@
...
+      }
+
+      if (self.localValidation) {
+        return callback(err, jwt);
+      }
+
+    });
+  }
+
+  jwksResolver(kid, callback) {
```
rchild-okta: Am curious - does njwt cache the keys?

lib/okta/access-token-authenticator.js
```diff
@@ -0,0 +1,87 @@
...
+    };
+    requestExecutor(req, function (err, res) {
+      if (err) {
+        return callback(err);
+      }
+      if (!res.keys || !res.keys.length) {
+        return callback('Keys collection not found');
+      }
+
+      const matches = res.keys.filter((key) => key.kid === kid);
```
rchild-okta: nit: I like find here instead of filter (will return immediately when it gets the match)

lib/okta/get-app-user-by-id.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+const requestExecutor = require('./request-executor');
+
+function getAppUserById(org, appId, apiToken, uid, callback) {
+  const req = {
+    url: org + '/api/v1/apps/' + appId + '/users/' + uid,
```
rchild-okta: Heh, as an aside, when moving to es6 it's really handy to use template literals for things like this, i.e.
```javascript
url: `${org}/api/v1/apps/${appId}/users/${uid}`
```

lib/okta/password-grant.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+var xtend = require('xtend');
+
+var requestExecutor = require('./request-executor');
+
+function passwordGrant(config, userFormSubmission, callback) {
+  var req = {
+    url: config.okta.org + 'oauth2/' + config.okta.authorizationServerId + '/v1/token',
```
rchild-okta: Although if we change the example to not show a trailing '/', then this would need to change. I'm okay with either way, but think we should be consistent (or when processing the config, normalize it).

okta-todo.md
```diff
@@ -0,0 +1,37 @@
...
+[ ] Email verification
+[ ] Password reset
+[ ] Client credentials authentication w/ keys as app user credentials
+[ ] Remove dependencies on Stormpath configuration
+[ ] Try to get current tests working, or rely on TCK?
+
+Todo tasks (discovered while implemented Primary goals):
+
+[ ] ensure that the migrated AS configuration will have the right settings for access token timeouts
+[ ] caching of jwks (can use HTTP response from .well-known)
```
rchild-okta: Ah, I see, nm about the caching question

lib/helpers/authenticate.js
```diff
@@ -28,42 +28,45 @@ module.exports = function authenticate(options, req, res, callback) {
...
                 return callback(err);
               }
 
-              callback(null, expandedAccount, authResult);
+              callback(null, user, oauthAccessTokenResult);
             });
           }
 
-          callback(null, expandedAccount, authResult);
+          callback(null, user, oauthAccessTokenResult);
```
rchild-okta: Just double checking - are all the objects returned to the user of the same format? I.e. in this case is oauthAccessTokenResult the same as authResult in shape (is this even publicly exposed)?

lib/okta/password-grant.js
robertjd: Let's normalize it, I'll add to this PR.

lib/middleware/get-user.js
robertjd: Right.  This is a strange check, I'm not sure why it was here in the first place.  In this context the authentication attempt has already happened, which would have failed if the user were not enabled.

lib/helpers/authenticate.js
robertjd: Ah, good catch, thanks!  I do need to mat this match the [https://docs.stormpath.com/nodejs/jsdoc/OAuthPasswordGrantAuthenticationResult.html](OAuthPasswordGrantAuthenticationResult) as much as possible.

PR #612, opened 2017-04-18
next URI must be encoded in the form action for query strings to work…

bosticka:
… on redirect.

formAction is not encoded.  So if your next URI has a query string, its then a query string in side a query string and breaks.  Wrapping in encodeURIComponent fixes.  Please pull.


robertjd:
Thanks @bosticka !  We're busy building the next 4.x version, but I will review this as soon as possible.

PR #618, opened 2017-06-15
Fixed get user auth priority - first check cookie and then header for…

alexpx:
Migration from Stormpath to Okta:

When checking authenticationRequired route the resolvedAccessToken should take the access token from cookie first.

This scenario is happen when the request contains the cookies together with Auth Bearer and then the auth failed


robertjd:
Hi Alex, can you tell me the use case for sending both the cookie and the header?  Since this technically isn't part of the migration work I'm inclined to not change this current behavior, especially since we don't have a test for this case, thus I'm not sure who is depending on this current behavior.  A workaround would be to put your own middleware in front of this, and remove the token that you don't want to use.

